Issue #38, opened 2014-05-05
How to get NH to use an existing IDbTransaction


johnsimons:
So the issue is:
- SQLServer transport creates an `IDbConnection` and also an `IDbTransaction` 
- Those instances remain open for the duration of the request
- NHibernate can easily use the `IDbConnection` but it doesn't look like it can use the existing `IDbTransaction`, see https://nhibernate.jira.com/browse/NH-1855

I actually don't know for sure if we even need to share the same transaction as the SQLServer transport or not?

@andreasohlund @johannesg thoughts?

johnsimons:
We have decided to throw a better exception

SzymonPobiega:
There is one big advantage actually. One cannot use Outbox when using ambient transactions. This means one cannot use SQLTransport multi-db feature without DTC. I'll take a look at `ITransactionFactory` again.

SzymonPobiega:
I did a small spike in order to be absolutely sure we can/cannot do this and it seems like it is very unlikely that without resorting to High Magic (passing arguments to a method via thread-static objects) we can do this.

I did implement an `ITransactionFactory` that uses an existing `IDbTransaction` and I would be able to expose session with such a transaction factory to the user but because of the fact that this session has been constructed using an existing connection, it would (silently) not support _2nd level cache_ despite all the correct config etc. The bloker is (this line](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Impl/SessionFactoryImpl.cs#L477)

I proved that Outbox can be enabled without the need to suppress ambient transactions so closing this.

PR #77, opened 2015-01-22
NHibernate persistence and SQL transport do not work well together without TransactionScope

particularbot:
### Symptoms

When looking for a saga instance an exception is thrown:
```
System.InvalidOperationException: SqlConnection does not support parallel transactions.
```

### Who's affected

Users who use both SQLServer transport and NHibernate peristence but with disabled usage of `TransactionScope`:

```
configuration.Transactions().Enable();
configuration.Transactions().DisableDistributedTransactions();
```

### Workaround

Do not disable usage of `TransactionScope`

### Possible solution

As described [here](https://github.com/Particular/NServiceBus.NHibernate/issues/38#issuecomment-65210350) the solution might be to implement a custom `ITransactionFactory`. Alternatively we need to throw a more descriptive exception.


SzymonPobiega:
Added a better exception

SzymonPobiega:
@johnsimons can you review?

johnsimons:
@SzymonPobiega do you need help finishing this one off ?

SzymonPobiega:
@johnsimons I've added a startup-time check using a feature that depends on both SQL transport and the shared storage. Do you think it makes sense?

SzymonPobiega:
@johnsimons hold on with merging this for a moment

SzymonPobiega:
Sorry for that. I wanted to do a small spike to absolutely sore we can't do this and now I am absolutely sure. More details in #38

johnsimons:
@SzymonPobiega before merging, can you address https://github.com/Particular/NServiceBus.NHibernate/commit/3f6f09832d48699198f5d87c53450b165e9b9ce4#comments

SzymonPobiega:
Hmmm.. I adressed it already... strange. Also 278 changed files seems wrong. Will push -f

SzymonPobiega:
Files are OK, these are acceptance tests that I upgraded. The exception text is OK as per last commit ('The transaction WILL NOT be escalated bla bla ble')

johnsimons:
:+1: to merge

PR #79, opened 2015-02-17
Automatically disable reusing of transport connection when SQL Server transport is detected and Outbox is on

SzymonPobiega:
This pull adds a config option to disable sharing the transport connection by the persistence.


src/NServiceBus.NHibernate/SharedSession/SharedSessionConfig.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// <summary>
+    /// Deduplication configuration extensions.
+    /// </summary>
+    public static class SharedSessionConfig
+    {
+        /// <summary>
+        /// Disables sharing a connection between SQL Server transport and NHibernate persistence. Intended to be used in conjunction with Outbox only.
+        /// </summary>
+        /// <param name="persistenceConfiguration"></param>
+        public static PersistenceExtentions<NHibernatePersistence> DoNotShareTransportConnection(this PersistenceExtentions<NHibernatePersistence> persistenceConfiguration)
```
johnsimons: Very confusing API IMO.
When would a user need to turn this on/off ?
You mention "Intended to be used in conjunction with Outbox only.", the api as it stands does not give any clue apart from the api doco, that is the intent.

Maybe we should consider making this setting part of the Outbox API configuration, eg:
```c#
config.UsePersistence<NHibernate, StorageType.Outbox>()
                                                                .DoNotShareTransportConnection();
```

SimonCropp: unless we are mutating the return to a different type I think it should return void.

johnsimons: No the return type is correct @SimonCropp.
The `void` would break the chaining of the fluent API.

johnsimons: Still think calling this method `DoNotShareTransportConnection`, is not conveying intent, @SzymonPobiega can you explain in plain english what the intent is and then maybe we can name this appropriately ?

src/NServiceBus.NHibernate/SharedSession/SharedSessionConfig.cs
```diff
@@ -0,0 +1,21 @@
+namespace NServiceBus.Persistence.NHibernate
+{
+    using Configuration.AdvanceExtensibility;
+    using NServiceBus.Features;
+
+    /// <summary>
+    /// Deduplication configuration extensions.
```
johnsimons: I think this api doco is wrong, but I don't think it matters anyway.

src/NServiceBus.NHibernate/SharedSession/SharedSessionConfig.cs
SimonCropp: > The void would break the chaining of the fluent API.

My point exactly. Fluent APIs should have terminating methods. Chaining something off `DoNotShareTransportConnection` implies you are mutating the result of `DoNotShareTransportConnection` and that order is important, neither of which is correct

johnsimons:
Just wondering if it wouldn't be simple to replace the impl of `IDbConnectionProvider` instead ?
And then we can reuse `StorageSessionProvider`

src/NServiceBus.NHibernate/SharedSession/SharedSessionConfig.cs
johnsimons: That is not the intent @SimonCropp.
The chaining in this API is important and there is no order.
In this API the user can do things like:
```c#
config.UsePersistence<NHibernate, StorageType.Outbox>()
                                     .DoNotShareTransportConnection()
                                     .ConnectionString(connString)
                                     .ToggleSomeOtherConfigForOutbox();
```
And the above are all the configuration settings for `NHibernate` and `StorageType.Outbox`.
Does this make sense ?

SzymonPobiega: The main idea behind this setting is to be able to use Outbox for SQL Server transport, In such case we should not share the transport connection. Instead, we're using a `TransactionScope` created by `OutboxDeduplicationBehavior` to guarantee that the changes done by user, in saga storage and outbox storage are atomic. We're relying on the fact that each of these storages opens its own session within that `TransactionScope` but since these sessions are short-lived, they do not overlap with each other (which would cause DTC escalation).

That said, I *think* (not 100% sure) we can use the same approach for non-SQL Server transports + outbox configurations and get rid of `NHibernateStorageContext`.

johnsimons: > The main idea behind this setting is to be able to use Outbox for SQL Server transport

So if the user does not configure this setting, they won't be able to use it ?

SzymonPobiega:
I don't think so. The `ExecuteInTransaction` is the main reason I did it here. For the non-shared mode it creates a new session each time.

src/NServiceBus.NHibernate/SharedSession/SharedSessionConfig.cs
SzymonPobiega: That's partially correct. `.EnableOutbox()` disables the ambient transactions and user would get an exception when starting. On the other hand, If he calls `EnableFeature<Outbox>()` that would not disable the ambient transactions. In such case to properly handle updates user needs to use `NHibernateStorageContext` to access his data. If he creates a session himself (which e.g. is required in order to use second-level cache), that session's connection will be enlisted in `OutboxDedupBehavior` new transaction scope, not the transport one resulting in potential inconsistent operations.

SzymonPobiega:
After building a [transaction type/transport/outbox matrix](https://docs.google.com/a/nservicebus.com/spreadsheets/d/1dn-l1QPi-27yShxmHX4VY-M8dwIkbaxWT2h80YWCZB4/edit?usp=sharing) I am convinced that we should automatically disable connection sharing when outbox is on. We don't need the explicit API for that.

johnsimons:
Sorry @SzymonPobiega I can;t really provide advice here, I don't understand the matrix!

SzymonPobiega:
Focusing on SQL-related part, there are two cases in which I want to disable to transport connection sharing:
 * When DTC is off and Outbox is on. Currently we throw various kinds of 'concurrent transactions / no transactions' exceptions when user tries this so
 * When DTC is on and Outbox is on. This is not supported case where user uses `EnableFeature<Outbox>()` instead of `EnableOutbox()`, In this case the behavior was wrong since user data could be stored in Outbox TransactionScope instead of the transport one if user has not used the shared context.
I don't see any degradation of usability for users who are on the happy path (shared connection, no outbox). Are you fine with merging?

ramonsmits:
@SzymonPobiega Why wouldn't we want to share the connection when DTC is off and outbox is on?

I assume that when I have outbox and SQL persistence that these do share the connection?

Do I understand correctly that we currently do not support nodtc, outbox, sql transport, sql persistence combo?

If the connection is not shared, will FLR / SLR work when processing fails and transactions are enabled?

johnsimons:
@SzymonPobiega having a hard time evaluating the impact of this change, would you be able to describe a bit more about that ?
What does the public Api looks like, can u give an example ?
Please update the description of the PR

SzymonPobiega:
@johnsimons there is no change in the public API now. The impact is limited to the cases which were throwing errors before the change so no existing working deployments should be affected. The change does only matter for the new SQL transport release.

I am taking care of the release announcements.

johnsimons:
:+1: to merge

SzymonPobiega:
@johnsimons what do you think about a minor release containing this?

PR #89, opened 2015-03-09
Fixes breaking change with missing NHibernateStorageContext when using outbox

### LINKED ISSUE #86: When using Outbox user data access transaction completes even though there was an exception in the handler
### Symptoms

User data updates are executed multiple times in case an exception happens in the handler.

### Who's affected

All users who enable Outbox except those who run latest SQL Server transport (2.1) with NHibernate 6.1

### Root cause

This is because `TryGet` from outbox store in the `OutboxDeduplicationBehavior` is accessing `LazySqlConnection` causing a new connection to be created before the `TransactionScope` is started and that connection is reused for user data (and saga data).

### Proposed solution

Disable connection sharing in the `TryGet` method.

SzymonPobiega:
6.1.0 contained an unintended breaking change that affected people who use outbox with transports other than SQL Server (mainly RabbitMQ). The `NHibernateStorageContext` was missing from the container in the outbox mode. This is now fixed.
 * `TryGet` in `OutboxPersister` now uses NH's owned connection and explicitly uses NH's transactions since it is designed to be used outside of ambient transaction scope. This prevents the shared connection to be opened before the `OutboxDeduplicationBehavior` begins its ambient transaction (fixed #86).
 * Manual connection management and connection sharing between saga store and outbox store are enabled by default. This prevents [NH-3023](https://nhibernate.jira.com/browse/NH-3023). NB [NH doco](http://nhibernate.info/doc/nh/en/#transactions-connection-release) lies about releasing connections in context of an ambient transaction claiming it is done after each statement.
 * `NHibernateStorageContext` is exposed by default
 * When outbox is enabled for SQL Server transport `NHibernateStorageContext` is not exposed, connection management is delegated to NHibernate and no connection sharing is used. NHibernate native transactions are used always when accessing the database. It has been proved that for SQL Server these native transactions work correctly when used inside the `TransactionScope`. It has also been proved that SQL Server driver, when in single phase mode, does not run the `TransactionScope` events on a separate thread which makes it immune to [NH-3023](https://nhibernate.jira.com/browse/NH-3023).


src/NServiceBus.NHibernate/Outbox/OutboxPersister.cs
```diff
@@ -19,9 +20,9 @@ public bool TryGet(string messageId, out OutboxMessage message)
 
             message = null;
 
-            using (var session = StorageSessionProvider.OpenStatelessSession())
+            using (var session = SessionFactoryProvider.SessionFactory.OpenStatelessSession())
             {
-                using (var tx = session.BeginAmbientTransactionAware(IsolationLevel.ReadCommitted))
+                using (var tx = session.BeginTransaction(IsolationLevel.ReadCommitted))
```
SzymonPobiega: This is always called outside of `TransactionScope`

src/NServiceBus.NHibernate/Outbox/OutboxPersister.cs
```diff
@@ -19,9 +20,9 @@ public bool TryGet(string messageId, out OutboxMessage message)
 
             message = null;
 
-            using (var session = StorageSessionProvider.OpenStatelessSession())
+            using (var session = SessionFactoryProvider.SessionFactory.OpenStatelessSession())
```
SzymonPobiega: This thing opens a new session with NH's managed connection to fix #86

src/NServiceBus.NHibernate/SharedSession/NonSharedConnectionStorageSessionProvider.cs
```diff
@@ -12,16 +12,6 @@ public NonSharedConnectionStorageSessionProvider(SessionFactoryProvider sessionF
             this.sessionFactoryProvider = sessionFactoryProvider;
         }
 
-        public IStatelessSession OpenStatelessSession()
-        {
-            return sessionFactoryProvider.SessionFactory.OpenStatelessSession();
-        }
-
-        public ISession OpenSession()
```
SzymonPobiega: This thing had one usage which could be replaced by `OpenStatelessSession()`

src/NServiceBus.NHibernate/SharedSession/NonSharedConnectionStorageSessionProvider.cs
```diff
@@ -12,16 +12,6 @@ public NonSharedConnectionStorageSessionProvider(SessionFactoryProvider sessionF
             this.sessionFactoryProvider = sessionFactoryProvider;
         }
 
-        public IStatelessSession OpenStatelessSession()
```
SzymonPobiega: Since I disabled sharing of connection in this session, I could remove that and uses the `ISessionFactory` directly.

SzymonPobiega:
@johnsimons @andreasohlund @janovesk can you review?

src/NServiceBus.NHibernate/Outbox/OutboxPersister.cs
andreasohlund: Is that always true? eg ASB calls it with a scope right now? (which is wrong but still)  Should we suppress to make sure?

or should the outbox behavior in the core suppress?

SzymonPobiega: You're right, I think we should explicitly suppress here. Even better, we should suppress in the core (`OutboxDeduplicationBehavior`)

johnsimons:
Sorry @SzymonPobiega I have no idea what this code does.
Maybe this whole thing is way too complex.
I can review it with you so u can explain what is going on ?

SzymonPobiega:
@johnsimons how about a zoom session Thursday morning?

src/NServiceBus.NHibernate/SqlServerTransport/ValidateAmbientTransactionOrOutbox.cs
```diff
@@ -0,0 +1,30 @@
...
+            DependsOnAtLeastOne("NHibernateOutboxStorage", "SqlServerTransport"); 
+        }
+
+        protected override void Setup(FeatureConfigurationContext context)
+        {
+            if (context.Settings.GetOrDefault<bool>("Transactions.SuppressDistributedTransactions") 
+                && !context.Container.HasComponent<OutboxPersister>())
+            {
+                throw new InvalidOperationException(@"In order for NHibernate persistence to work with SQLServer transport, either ambient transactions or outbox need to be enabled. 
+In any case, the transaction WILL NOT be escalated to a distrubuted transaction if transport, NServiceBus persistence and user persistence all use exactly the same connection string.");
```
johnsimons: spelling mistake "distrubuted"

src/NServiceBus.NHibernate/SharedSession/NHibernateStorageSession.cs
```diff
@@ -79,5 +77,26 @@ protected override void Setup(FeatureConfigurationContext context)
             Installer.RunInstaller = context.Settings.Get<bool>("NHibernate.Common.AutoUpdateSchema");
             Installer.configuration = configuration;
         }
+
+        static bool DisableConnectionSharing(FeatureConfigurationContext context, Configuration configuration)
+        {
+            return context.Settings.GetOrDefault<bool>("NServiceBus.Features.SqlServerTransportFeature")
+                   && context.Settings.GetOrDefault<bool>(typeof(Outbox).FullName)
+                   && (SqlServerDriver(configuration) || SqlServerDialect(configuration));
```
johnsimons: I really dislike this, this is going to be a maintenance hole!
Why are we doing this ?

SzymonPobiega: I don't like it either. This makes the SQL/NH/Outbox mode not require `NHibernateStorageContext` in order for user to store data.

SzymonPobiega: I've raised a general concern regarding how we should address that https://github.com/Particular/Requirements/issues/258

johnsimons:
@SzymonPobiega @indualagarsamy based on https://nservicebus.desk.com/agent/case/10327, I think we need to get this out ASAP, this is message loss as far as I understand, do you agree ?

SzymonPobiega:
Question is, should we just cherry pick the commit that fixes the connection behavior? The user is also complaining about the NHibernateStorageContext being null (not causing message loss) and that second thing requires 'full' 6.1.1

We still have some randomness in the tests: http://builds.particular.net/viewType.html?buildTypeId=NServiceBusNHibernate_AcceptanceTestsSqlServer&branch_NServiceBusNHibernate=hotfix-6.1.1&tab=buildTypeStatusDiv. I've killed two that were failing and now I got another two.

PR #91, opened 2015-03-30
Updating to v6 core

johnsimons:


johnsimons:
@andreasohlund @danielmarbach this PR updates NH to use the core next major code.

I have a go at modifying the `SagaModelMapper` to consume the new exposed `SagaMetaModel` from the core, but so far I haven't come across any benefits.
Can one of you guys see if you find benefits on using  `SagaMetaModel` in NH ?

johnsimons:
@andreasohlund did more digging into updating the `SagaModelMapper` to use `SagaMetaModel`.
As we discussed yesterday we were not sure why we need https://github.com/Particular/NServiceBus.NHibernate/blob/e117383fb575a1215f380f8e467e6849e588aa6b/src/NServiceBus.NHibernate/SagaPersisters/SagaModelMapper.cs#L280-L292

After looking at blames, I figured that the first loop `var derivedTypes = typesToScan.Where(t => t.IsSubclassOf(rootEntity));` is needed to ensure properties base types are also mapped, eg:
```c#
public class MySagaData : ContainSagaData
{
    [Unique]
    public virtual Guid DataId { get; set; }

    public virtual PolymorpicPropertyBase PolymorpicRelatedProperty { get; set; }
}

public class PolymorpicProperty : PolymorpicPropertyBase
{
    public virtual int SomeInt { get; set; }
}

public class PolymorpicPropertyBase
{
    public virtual string SomeString { get; set; }
}
``` 

However the `var superClasses = typesToScan.Where(t => t.IsAssignableFrom(rootEntity));`, I haven't figure out why you added it in https://github.com/Particular/NServiceBus/commit/a4ec25eda28e9fd8b58421c71fac85a67dd31624#diff-6db26b509961f5a519e89b43b8b6a26fR207 ?

DavidBoike:
@johnsimons I swear I pinged you someplace a few days ago but can't find it so I may not have. Who knows.

Did some work to get NSB.NH going with the newest V6 but ran into some refactorings with SagaMetadata in some tests that Andreas said you should look at. Any advice?

johnsimons:
@DavidBoike what can I help with ?

DavidBoike:
@andreasohlund thought that perhaps the SagaMetadata stuff was your WIP. If that's not the case then I can take another stab at it.

Issue #96, opened 2015-05-06
Installers don't work properly when hosting multiple endpoints in a single AppDomain


SzymonPobiega:
### Symptoms

The `Configuration` object in the installer gets overwritten by other endpoints hosted in the same AppDomain because it is set via a static field which results in only one endpoint being able to automatically create tables.

### Who's affected

User's of NHibernate-based persistence who wish to host more than one endpoint in a single AppDomain.

### Workaround

Here's possible workaround https://github.com/AlanMullett/NServiceBus.NHibernate/tree/Branch-6.1.0-multithread.

SzymonPobiega:
Another solution would be to use containers for installers: https://github.com/Particular/NServiceBus.NHibernate/pull/97. @johnsimons can you take a look? I've committed an acceptance test but it does not prove anything since in ATs endpoints are activated sequentially so the static does not cause any problems.

PR #97, opened 2015-05-06
Multi host enabled installers

SzymonPobiega:
Attempts to solve #96 by registering the installers in the container


SzymonPobiega:
@Particular/engineers please review. I've checked running the installers via NSB.Host -install and work OK

andreasohlund:
Looks good to me

PR #98, opened 2015-05-08
Configuration file is validated even though user explicitly overridden it by passing his own `Cfg.Configuration` object

SzymonPobiega:
### Symptoms

When passing a `Configuration` object to the NHibernate persistence configurator the user has to specify the connection string twice:
 * in the `Configuration` object
 * by calling `.ConnectionString` on NHibernate peristence config (or using a naming convention)

### Who's affected

User's who wish to provide their own NH config, e.g. because they want to override the default schema (dbo)

### Root cause

NH persistence features do validate the settings passed in both code and app.config even though the user has explicitly overridden them by passing his `Configuration` object.


SzymonPobiega:
@Particular/engineers can you review? Please use the commits tab to review the actual commit. The other is updating the ATs so it is HUGE.

PR #113, opened 2015-07-28
Lock index and query improvements

ramonsmits:
This PR has a few optimizations in regards to timeout data.
## Changes
- Using update lock in code that will always do a delete after the select. This will prevent deadlocks.
- Property order rearranged so that SQL index 'TimeoutEntity_EndpointIdx' will have correct column ordering as this cannot be specified in the mapping.
- Query adjusted so that it can be completely performed on the non-clustered index (no need to use clustered index to retrieve resultset values)
- Using NHibernate native guidcomb instead of the copied implementation.
## Issues
- The index column ordering should be communicated as this needs to be resolved manually. The index will not be updated by NHibernate.
## Tasks left
- [ ] Review subscription entity
- [ ] Review outbox entity
- [x] Rebase on `develop` branch


SzymonPobiega:
@ramonsmits why would going with `Endpoint, Time` be better than `Time, Endpoint` for the index here?

ramonsmits:
The reason for endpoint and then time is that data on disk is read more efficiently when doing the GetNextCheck query. This queries on:

* Endpoint equal
* Time between
* order by time


Imagine lots of timeouts of lots of endpoints. If time is first and you do an index seek and start reading you would hit data of other endpoints.

By using this order SQL can do a non fragmented read until the end time. All rows read will be valid and do not need to be filtered.

Its a small optimisation but that is the reason for the change.

ramonsmits:
@SzymonPobiega I've updated this PR with develop. Please let me know if you understand the changes. I'll assign you in the mean time.

SzymonPobiega:
@Particular/persistence-maintainers should we consider it for V6?

SimonCropp:
@Particular/nhibernate-persistence-maintainers @ramonsmits so is this abandoned? should we delete?

cremor:
Please don't delete this pull request. The points are still valid for v7 (core v6).

The query adjustment might need reevaluation in v7 because now the Id is needed in addtion to the time. I think (but didn't test) that in MS SQL Server that would still not require a clustered index seek but other databases might be different.

Changes for the Guid generation are even more important in v7 because of #260

SzymonPobiega:
@Particular/nhibernate-persistence-maintainers are we going to do anything with this?

MarcinHoppe:
I'd vote to merge when the build is green.

SzymonPobiega:
Replaced by https://github.com/Particular/NServiceBus.NHibernate/pull/318

PR #120, opened 2015-09-22
implement IPersistTimeoutsV2 interface

timbussmann:
Implement the IPersistTimeoutsV2 interface in order to fix loosing timeout messages when not using dtc.

see issue https://github.com/Particular/NServiceBus/issues/2885

Connects to Particular/NServiceBus#2885

### DIFF
--- a/src/NServiceBus.NHibernate.Tests/TimeoutPersister/When_fetching_timeouts_from_storage.cs
+++ b/src/NServiceBus.NHibernate.Tests/TimeoutPersister/When_fetching_timeouts_from_storage.cs
@@ -11,7 +11,7 @@ namespace NServiceBus.TimeoutPersisters.NHibernate.Tests
     public class When_fetching_timeouts_from_storage : InMemoryDBFixture
     {
         [Test]
-        public void Should_return_the_complete_list_of_timeouts()
+        public void GetNextChunk_should_return_the_complete_list_of_timeouts()
         {
             const int numberOfTimeoutsToAdd = 10;
 
@@ -33,7 +33,7 @@ namespace NServiceBus.TimeoutPersisters.NHibernate.Tests
         }
 
         [Test]
-        public void Should_return_the_next_time_of_retrieval()
+        public void GetNextChunk_should_return_the_next_time_of_retrieval()
         {
             var nextTime = DateTime.UtcNow.AddHours(1);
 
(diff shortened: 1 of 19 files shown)


timbussmann:
@Particular/nservicebus nhibernate experts please review

ramonsmits:
As discussed during the review:

* Normal operation is peek, send, delete. During peek we can do a row lock so that the delete will always succeed. Currently the peek and delete are executed in the same transaction when using connection sharing between sql transport and nhibernate persistence. Maybe this can potentially happen when running with multiple instances are fetching timeouts as these instance are competing on the timeout queue. (`ISession.Get<T>(object, LockMode.Upgrade`)
* The entity mapping is duplicate except for one property. Unknown why this is or what behavioral change this will have. Maybe refactor to one mapping due to the number of properties and using object initializer instead of a constructor which can potentially result in missing properties if the object is extended in the future.

timbussmann:
@ramonsmits just had a look at the `LockMode.Upgrade` when peeking but I get some unwanted behavior in the unit test. If you have two concurrent transactions creating an update lock, can they deadlock each other?

ramonsmits:
No, that should not be the case as yo do a SELECT FOR UPDATE, that should not result in a deadlock, it can only result in a sequential operation. 2nd transaction will have to wait before the first completes but the 1st does not have to wait for the 2nd because it already acquired a row write lock.

timbussmann:
@ramonsmits never mind, It seems to be an issue with SQLite, same code works on SQLExpress.

timbussmann:
@ramonsmits please review the latest commit, I changed the code according to your points:
* using upgrade lock for peek. This should only improve scenarios when using dtc for now, since as we saw, this version doesn't yet support connection sharing. So the only case where that lock may be held until the TryRemove is for ambient transactions.
* extracted the entity mapping so both `Peek` and `TryRemove` behave the same.

src/NServiceBus.NHibernate.Tests/TimeoutPersister/When_fetching_timeouts_from_storage.cs
```diff
@@ -11,7 +11,7 @@ namespace NServiceBus.TimeoutPersisters.NHibernate.Tests
     public class When_fetching_timeouts_from_storage : InMemoryDBFixture
     {
         [Test]
-        public void Should_return_the_complete_list_of_timeouts()
+        public void GetNextChunk_should_return_the_complete_list_of_timeouts()
```
ramonsmits: Why the method name change? Test is unaltered?

src/NServiceBus.NHibernate.Tests/TimeoutPersister/When_removing_timeouts_from_the_storage.cs
```diff
@@ -79,30 +81,97 @@ public void Should_remove_timeouts_by_id()
...
+
+            Assert.IsTrue(t1Result.HasValue && t2Result.HasValue);
+
+            // one delete should succeed, the other one shouldn't
+            Assert.IsTrue(t1Result.Value || t2Result.Value);
+            Assert.IsFalse(t1Result.Value && t2Result.Value);
+        }
+
+        [Test]
+        public void RemiveTimeoutBy_should_remove_timeouts_by_sagaid()
```
ramonsmits: Typo in method

src/NServiceBus.NHibernate.Tests/TimeoutPersister/When_removing_timeouts_from_the_storage.cs
```diff
@@ -79,30 +81,97 @@ public void Should_remove_timeouts_by_id()
...
 
         [Test]
-        public void Should_remove_timeouts_by_sagaid()
+        public void TryRemove_should_return_false_when_timeout_already_deleted()
+        {
+            var timeout = new TimeoutData();
+
+            persister.Add(timeout);
+
+            Assert.IsTrue(persister.TryRemove(timeout.Id));
```
ramonsmits: How is this even possible? The timeout being stored has no ID assigned. Is it assigned by the database?

src/NServiceBus.NHibernate.Tests/TimeoutPersister/When_removing_timeouts_from_the_storage.cs
```diff
@@ -79,30 +81,97 @@ public void Should_remove_timeouts_by_id()
...
+            Assert.IsFalse(persister.TryRemove(timeout.Id));
+        }
+
+        [Test]
+        public void TryRemove_should_work_with_concurrent_transactions()
+        {
+            var timeout = new TimeoutData()
+            {
+                Id = Guid.NewGuid().ToString(),
```
ramonsmits: Conflicting behavior regarding ID assignment with test `TryRemove_should_return_false_when_timeout_already_deleted `

src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutStorage.cs
```diff
@@ -189,6 +180,47 @@ static string ConvertDictionaryToString(ICollection data)
             return serializer.SerializeObject(data);
         }
 
+        public TimeoutData Peek(string timeoutId)
```
ramonsmits: As mentioned, I don't agree with this method name. Its an entity (get, delete) and not a message (peek/receive). Maybe an improvement for v6.

src/NServiceBus.NHibernate.Tests/TimeoutPersister/When_removing_timeouts_from_the_storage.cs
timbussmann: yes, timeout id is always assigned by the timeout persister

src/NServiceBus.NHibernate.Tests/TimeoutPersister/When_fetching_timeouts_from_storage.cs
timbussmann: to distinct between `GetNextChunk` and `Peek` tests

src/NServiceBus.NHibernate.Tests/TimeoutPersister/When_removing_timeouts_from_the_storage.cs
timbussmann: oh that id assignment shouldn't be there. Thats some remainder from the SQLite tests. Going to remove that.

src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutStorage.cs
timbussmann: happy to discuss this on the v6 solution.

timbussmann:
@Particular/nservicebus @ramonsmits any further concerns or can we merge this?

andreasohlund:
Looks good to me

PR #150, opened 2016-01-28
Fixing oracle tests

HEskandari:
Addressing issues in oracle tests, brittle tests and hard-coded driver


SimonCropp:
@HEskandari if this works u r awesome

HEskandari:
There are other things to do as well...without this, it wouldn't even run locally for me (dll hell)!

SzymonPobiega:
@HEskandari can you elaborate on switching to the managed driver? Is that something our users should be concerned with?

HEskandari:
@SzymonPobiega I don't see how this relates the users? This is done on the acceptance test for the NHibernate persistence, not the library itself. The reason for using the managed driver:

 - It is all managed code, no dependencies to native oracle dll is needed.
 - pulled down via nuget, as opposed to stored dll in the source code.
 - has smaller footprint (a few mb vs. downloading a 100mb package)
 - the stored dll didn't work with my Oracle 11g express to begin with (the managed driver did work)

The managed driver, in short, should be preferred for its similicity: you use one dll to communicate with the database from your code and don't need any external packages/dependencies.

SzymonPobiega:
@HEskandari ah, ok. Makes perfect sense. I didn't notice the fact that the change only applies to the acceptance tests. Sorry. My fault.

HEskandari:
@SzymonPobiega not a problem!

SimonCropp:
@HEskandari this looks all good to me. anything left to be done?

HEskandari:
@SimonCropp there's still a couple of tests failing...trying to fix as much as possible before merge this in.

HEskandari:
There is a know (not sure if documented though) issue with using Oracle + NHibernate with a "StringClob" type (for large texts), which is causing the "When_using_hbms->Should_use_saga_entity_hbm" test to fail. I have tried a couple of things to work my way around it, but couldn't get it to work.

What's the consensus? can we merge it in?

SzymonPobiega:
:+1: for merging.

PR #154, opened 2016-02-02
Fixed handling of transactions in Add and TryRemove methods.

SzymonPobiega:
Now these methods should properly recognize the transport transaction (be it TransactionScope or a TS+SqlConnection) and hook into it in order to provide exactly-once experience.


SzymonPobiega:
@weralabaj @HEskandari please review.

ramonsmits:
There is missing context here. Why is this needed? Where is the issue?

SzymonPobiega:
@ramonsmits as in description, the current code ignores the transport transaction and/or SQL connection which results in at-least-once behavior where we expect exactly-once (DTC MSMQ or SQL).

weralabaj:
@SzymonPobiega FYI I updated the package to this PR, but that doesn't seem to solve the timeout issue. I'll need more time to investigate and would be good to get the latest core, so I'll look into that on Monday.

SzymonPobiega:
Closing this one and creating another one based on develop.

SzymonPobiega:
Replaced by #160

Issue #164, opened 2016-02-24
Deferred messages not delivered until endpoint restart


tmasternak:
taskforce: @tmasternak @timbussmann 

For V6 related discussion see: #https://github.com/Particular/NServiceBus/issues/3500
## Who's affected
- Any user user who uses NServiceBus.NHibernate. Users of SQL Azure are more likely to experience the problem because of more frequent connection problems when using a remote database server.

## Symptoms

In certain rare condition deferred messages to be delivered only after endpoint restart. These conditions include connectivity problems with the underlying storage or performance problems of the storage itself, both cause the timeout save operation to not complete on time.

## Root cause

There is a race condition between saving message and updating timeout query window. If there are is a messages that takes long to get saved in the `TimeoutStore` and at the same time a message with larger `DeliverAt` gets dispatched then the first message will be delivered only after endpoint restarts.
## Plan of Attack
- [x] Create test which reproduces the issue
- [x] Create patch [PR](https://github.com/Particular/NServiceBus.NHibernate/pull/168)
- [x] Prepare announcement [PR](https://github.com/Particular/Announcements/pull/81)

tmasternak:
We have a test which reproduces the bug.

SzymonPobiega:
This has not been ported forward to 7. Re-opening

PR #166, opened 2016-02-25
TimeoutPersister makes a periodic query to make sure all due timeouts have been dispatched.

tmasternak:
Connects to #164

### DIFF
--- a/src/NServiceBus.NHibernate.AcceptanceTests/Issue_164.cs
+++ b/src/NServiceBus.NHibernate.AcceptanceTests/Issue_164.cs
@@ -1,0 +1,172 @@
+﻿namespace NServiceBus.NHibernate.AcceptanceTests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.Config;
+    using NServiceBus.Config.ConfigurationSource;
+    using NServiceBus.Features;
+    using NServiceBus.Persistence;
+    using NServiceBus.Timeout.Core;
+    using NUnit.Framework;
+
+    public class Issue_164 : NServiceBusAcceptanceTest
+    {
+        private static string CommandTagHeader = "Chaos.HeaderName";
+
+        [Test]
+        public void Subscribers_retry_with_SLR()
+        {
+            var ctx = new Context();
+
+            Scenario.Define(ctx)
+                .WithEndpoint<Endpoint>(b =>
+                    b.When(bus =>
+                    {
+                        ctx.MessageATag = "MessageA: " + Guid.NewGuid().ToString();
+                        ctx.MessageBTag = "MessageB: " + Guid.NewGuid().ToString();
+
+                        var deliverCommandAAt = DateTime.UtcNow.AddSeconds(5);
+                        var deliverCommandBAt = deliverCommandAAt.AddSeconds(1);
+
+                        var commandA = new Command();
+                        bus.SetMessageHeader(commandA, CommandTagHeader, ctx.MessageATag);
+                        bus.Defer(deliverCommandAAt, commandA);
+
+                        var commandB = new Command();
+                        bus.SetMessageHeader(commandB, CommandTagHeader, ctx.MessageBTag);
+                        bus.Defer(deliverCommandBAt, commandB);
+
+                    }))
+                .Done(c => c.MessageAReceived)
+                .Run();
+        }
+
+        public class Context : ScenarioContext
+        {
+            public bool MessageAReceived { get; set; }
+
+            public string MessageATag { get; set; }
+
+            public string MessageBTag { get; set; }
+        }
+
+        public class Endpoint : EndpointConfigurationBuilder
+        {
+            public Endpoint()
+            {
+                EndpointSetup<DefaultPublisher>(b =>
+                {
+                    b.UsePersistence<ChaosPersister>();
+                    b.PurgeOnStartup(true);
+                });
+            }
+
+            public class ProvideConfiguration : IProvideConfiguration<TransportConfig>
+            {
+                public TransportConfig GetConfiguration()
+                {
+                    return new TransportConfig
+                    {
+                        MaximumConcurrencyLevel = 10,
+                        MaximumMessageThroughputPerSecond = 10
+                    };
+                }
+            }
+
+            class Handler : IHandleMessages<Command>
+            {
+                public Context Context { get; set; }
+
+                public IBus Bus { get; set; }
+
+                public Context TestContext { get; set; }
+
+                public void Handle(Command message)
+                {
+                    if (Bus.GetMessageHeader(message, CommandTagHeader) == Context.MessageATag)
+                    {
+                        TestContext.MessageAReceived = true;
+                    }
+                }
+            }
+        }
+
+        [Serializable]
+        public class Command : ICommand
+        {
+        }
+
+        public class ChaosPersister : PersistenceDefinition
+        {
+            public ChaosPersister()
+            {
+                Supports(Storage.Timeouts, s => s.EnableFeatureByDefault<ChaosTimeoutStorage>());
+            }
+        }
+
+        public class ChaosTimeoutPersister : IPersistTimeouts
+        {
+            private ManualResetEvent waitWithMessageACommit = new ManualResetEvent(false);
+
+            public Context Context { get; set; }
+
+            public IPersistTimeoutsV2 OriginalPersisterV2 { get; set; }
+
+            private IPersistTimeouts OriginalPersister => (IPersistTimeouts) OriginalPersisterV2;
+
+            public IEnumerable<Tuple<string, DateTime>> GetNextChunk(DateTime startSlice, out DateTime nextTimeToRunQuery)
+            {
+                var result = OriginalPersister.GetNextChunk(startSlice, out nextTimeToRunQuery);
+
+                return result;
+            }
+
+            public void Add(TimeoutData timeout)
+            {
+                if (timeout.Headers[CommandTagHeader] == Context.MessageATag)
+                {
+                    waitWithMessageACommit.WaitOne();
+                }
+
+                OriginalPersister.Add(timeout);
+            }
+
+            public bool TryRemove(string timeoutId, out TimeoutData timeoutData)
+            {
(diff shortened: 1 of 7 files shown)


tmasternak:
@SzymonPobiega @timbussmann can you review?

src/NServiceBus.NHibernate.AcceptanceTests/Issue_164.cs
```diff
@@ -0,0 +1,187 @@
...
+    using NServiceBus.AcceptanceTests.EndpointTemplates;
+    using NServiceBus.Config;
+    using NServiceBus.Config.ConfigurationSource;
+    using NServiceBus.Configuration.AdvanceExtensibility;
+    using NServiceBus.Features;
+    using NServiceBus.Persistence;
+    using NServiceBus.Timeout.Core;
+    using NUnit.Framework;
+
+    public class Issue_164 : NServiceBusAcceptanceTest
```
timbussmann: I dislike naming test cases after issues.

src/NServiceBus.NHibernate.AcceptanceTests/Issue_164.cs
```diff
@@ -0,0 +1,187 @@
...
+    {
+        private static string CommandTagHeader = "TestTag.HeaderName";
+
+        [Test]
+        public void Timeout_does_not_get_stuck_when_persister_commit_takes_longer_than_dispatch_of_another_later_timeout()
+        {
+            //TODO: trick TM to query timeouts for frequently. Tweak GetNextChuns in adapter
+            //TODO: make clean-up interval and clean-up period configurable and override it in this test
```
timbussmann: what's with these todos?

src/NServiceBus.NHibernate.AcceptanceTests/Issue_164.cs
```diff
@@ -0,0 +1,187 @@
...
+            }
+
+            public class ProvideConfiguration : IProvideConfiguration<TransportConfig>
+            {
+                public TransportConfig GetConfiguration()
+                {
+                    return new TransportConfig
+                    {
+                        MaximumConcurrencyLevel = 10,
+                        MaximumMessageThroughputPerSecond = 10
```
timbussmann: do we need throttling?

src/NServiceBus.NHibernate.AcceptanceTests/Issue_164.cs
```diff
@@ -0,0 +1,187 @@
...
+            //TODO: trick TM to query timeouts for frequently. Tweak GetNextChuns in adapter
+            //TODO: make clean-up interval and clean-up period configurable and override it in this test
+            var ctx = new Context();
+
+            Scenario.Define(ctx)
+                .WithEndpoint<Endpoint>(b =>
+                    b.When(bus =>
+                    {
+                        ctx.CommandATag = "MessageA: " + Guid.NewGuid().ToString();
+                        ctx.CommandBTag = "MessageB: " + Guid.NewGuid().ToString();
```
timbussmann: wouldn't either tag or guid already suffice?

src/NServiceBus.NHibernate.AcceptanceTests/Issue_164.cs
```diff
@@ -0,0 +1,187 @@
...
+
+                        var messageADelay = TimeSpan.FromSeconds(5);
+                        var messageBDelay = TimeSpan.FromSeconds(6);
+
+                        var now = DateTime.UtcNow;
+                        var deliverCommandAAt = now.Add(messageADelay);
+                        var deliverCommandBAt = now.Add(messageBDelay);
```
timbussmann: I think it would be ok to inline `now.AddSeconds(x)`

src/NServiceBus.NHibernate/TimeoutPersisters/NHibernateTimeoutStorage.cs
```diff
@@ -54,10 +56,18 @@ protected override void Setup(FeatureConfigurationContext context)
                 }
             }
 
+            var timeoutsClenaupExecutionInterval =
```
timbussmann: Clenaup

timbussmann:
shouldn't that test fail the build?

src/NServiceBus.NHibernate.AcceptanceTests/Issue_164.cs
tmasternak: Those are left-overs. Removing.

src/NServiceBus.NHibernate.AcceptanceTests/Issue_164.cs
tmasternak: We do not. Removing.

src/NServiceBus.NHibernate.AcceptanceTests/Issue_164.cs
tmasternak: The tag is not enough. It is not easy to purge TM satellite queues and persister on startup. I'll make it guid only.

src/NServiceBus.NHibernate.AcceptanceTests/Issue_164.cs
tmasternak: :+1:

src/NServiceBus.NHibernate/TimeoutPersisters/NHibernateTimeoutStorage.cs
tmasternak: :+1:

src/NServiceBus.NHibernate.AcceptanceTests/Issue_164.cs
tmasternak: :+1:

tmasternak:
@timbussmann this includes not only reproduction but also fix. If you comment out [this](https://github.com/Particular/NServiceBus.NHibernate/pull/166/files#diff-d2cdc7c3c692226eca6d8057b7edb743R68) and next line which configure frequency of clean-up the test starts failing.

src/NServiceBus.NHibernate/TimeoutPersisters/NHibernateTimeoutStorage.cs
```diff
@@ -54,10 +56,18 @@ protected override void Setup(FeatureConfigurationContext context)
                 }
             }
 
+            var timeoutsCleanupExecutionInterval = 
+                context.Settings.GetOrDefault<TimeSpan?>("NHibernate.Timeouts.CleanupExecutionInterval") ?? TimeSpan.FromMinutes(2);
+
+            var timeoutsCleanupQueryPeriod =
+                context.Settings.GetOrDefault<TimeSpan?>("NHibernate.Timeouts.CleanupQueryPeriod") ?? TimeSpan.FromMinutes(5);
```
SzymonPobiega: Why do we need this? When the endpoint starts we query from 10 years before (AFAIR) so I guess here we can do the same thing?

tmasternak: Fair point. I did not want to make query to be 'from the beginning of time until now' but in fact we are doing this type of query on each endpoint start.

tmasternak:
@Particular/persistence-maintainers can you have a look?

timbussmann:
Looks good to me :+1:

src/NServiceBus.NHibernate/TimeoutPersisters/NHibernateTimeoutStorage.cs
```diff
@@ -54,10 +56,14 @@ protected override void Setup(FeatureConfigurationContext context)
                 }
             }
 
+            var timeoutsCleanupExecutionInterval = 
+                context.Settings.GetOrDefault<TimeSpan?>("NHibernate.Timeouts.CleanupExecutionInterval") ?? TimeSpan.FromMinutes(2);
```
SzymonPobiega: Should we have a public API for overriding the default here? Also, we should document why 2 minutes is a good default (and not 1 minute or 10 minutes). I've seen somebody rightfully recently challenged the "7 day rule" of outbox that we have no scientific data to back with.

tmasternak: Good point. Here are my arguments:
 * When it comes to default override we had a discussion with @andreasohlund and @timbussmann and decided that we will not expose the setting through api method for now. Thanks to that we: a. do not have to pump package major number to follow semver  and b. if any user will want to change that we can still give the guidance on how to do it. This change is very similar to RavenDB approach where far none of the users requested the ability to change the default.
 * When it comes to clean-up period once 2 min the arguments (not scientific for sure) are: 
  * this is value for RavenDB [QueryTimeouts](https://github.com/Particular/NServiceBus.RavenDB/blob/develop/src/NServiceBus.RavenDB/Timeouts/QueryTimeouts.cs#L19)
   * ~3 (2min from persister + 1min from poller) min is now worse case scenario time for which the deferred will be delayed. I do know it sounds pretty arbitrary but I would argue it should be a couple of minutes and probably no more than 10.

andreasohlund: :+1: so we are deferring the decision to introduce a public api but we have a workaround ready in the unlikely event of a user needing to change it.

SzymonPobiega: Yeah, like `unlikely event of losing cabin pressure` :smiley: Fine. Should we document that with NH persistence the worst case delay of timeout delivery when idle is 3 minutes? This is not a blocker for this pull. I am :+1: for beginning the hotfix procedure.

tmasternak: I'm not sure if we want to give any guarantees. I think the only thing we say about deferred messages is that they won't get delivered earlier than. @andreasohlund thoughts?

tmasternak:
Closing this one. Created proper hotfix branch https://github.com/Particular/NServiceBus.NHibernate/pull/168.

src/NServiceBus.NHibernate/TimeoutPersisters/NHibernateTimeoutStorage.cs
andreasohlund: > I think the only thing we say about deferred messages is that they won't get delivered earlier than. @andreasohlund thoughts?

I don't think we have any SLA for timeouts documented. Shouldn't we have that?

PR #171, opened 2016-03-03
Update to core 1616

SzymonPobiega:


SzymonPobiega:
@HEskandari seems like this is quite a good build - only one AT failed

PR #176, opened 2016-03-07
Added clearing of transport operation after dispatching

SzymonPobiega:
* Update the transport operations column with `NULL` when marking as dispatched
 * Added V5 patch that would skip returning transport operations if outbox record has been already dispatched


SzymonPobiega:
@ramonsmits @HEskandari please review.

src/NServiceBus.NHibernate/Outbox/OutboxPersister.cs
```diff
@@ -87,7 +91,7 @@ public Task SetAsDispatched(string messageId, ContextBag context)
                 {
                     using (var tx = session.BeginTransaction(IsolationLevel.ReadCommitted))
                     {
-                        var queryString = $"update {typeof(OutboxRecord)} set Dispatched = true, DispatchedAt = :date where MessageId IN ( :messageid, :qualifiedMessageId ) And Dispatched = false";
+                        var queryString = $"update {typeof(OutboxRecord)} set Dispatched = true, DispatchedAt = :date, TransportOperations = NULL where MessageId IN ( :messageid, :qualifiedMessageId ) And Dispatched = false";
```
DavidBoike: I know this is unchanged from previous, but should we be relying upon Type.ToString() here? For example, if this were made an internal nested class for some reason, it would go from `OutboxRecord` to `ParentClass+OutboxRecord`. Should we change this to use `{typeof(OutboxRecord).Name}` ?

(Probably in many other areas as well.)

DavidBoike:
Reviewed - I have one comment which isn't even directly related to your changes.

I'm :+1: to merge and will probably do so tomorrow if nobody else has objected.

HEskandari:
Looks good to me! :+1:

SzymonPobiega:
@HEskandari @DavidBoike I'll add `.Name` and ask you to merge.

SzymonPobiega:
@HEskandari @DavidBoike I'll add `.Name` and ask you to merge.

PR #179, opened 2016-04-01
Added API approval tests.

SzymonPobiega:
@Particular/persistence-maintainers please review.


SzymonPobiega:
The API spec change is in a separate commit

TL;DR
 * feature classes made internal
 * `NHibernateStorageContext` -> synchronized storage session

src/NServiceBus.NHibernate.Tests/API/APIApprovals.cs
```diff
@@ -0,0 +1,26 @@
...
+    using System.IO;
+    using System.Runtime.CompilerServices;
+    using ApiApprover;
+    using ApprovalTests;
+    using ApprovalTests.Reporters;
+    using Mono.Cecil;
+    using NUnit.Framework;
+
+    [TestFixture]
+    [Explicit("Marking it as explicit before we have all upgrade issue covered in a guide.")]
```
DavidBoike: What do you mean by this? I don't understand when this would come out and why.

DavidBoike:
@SzymonPobiega was the renamings to append Feature done as a consistency thing? Or to create alignment with the prior version so that the Obsoletes properly reflected the current state of things?

Looks good. And I think this should get merged first, and then we should think about publics/internals, so that the things getting removed from the public API will show up as a deletion from the public API txt file.

src/NServiceBus.NHibernate.Tests/API/APIApprovals.cs
SzymonPobiega: Sorry, this was a reminder to me that the tests were still breaking. Of course I forgot to remove it before pushing. I will fix it.

SzymonPobiega:
@DavidBoike yes, I added the `Feature` suffix to allow to distinguish the old publish features from new internal ones.

I've removed that leftover `obsolete` attribute.

SzymonPobiega:
@DavidBoike can you merge if OK?

DavidBoike:
I'm :+1: but it has conflicts now. Once you have conflicts sorted out go ahead and self-merge.

PR #186, opened 2016-04-21
Fix example syntax for Oracle-style lock

bbrandt:


SzymonPobiega:
Looks good. Thanks @bbrandt

PR #199, opened 2016-06-08
Fix connection sharing with transport

SzymonPobiega:
Removed the core responsible for sharing an instance of `SqlConnection`. Based on the changes done in https://github.com/Particular/NServiceBus.SqlServer/pull/252 the same goal will be achieved by having separate instances of `SqlConnection` in transport and in persistence provided that both use the same connection string. The key feature is early closing of transport connection that prevents DTC escalation when persistence connection is created.


SzymonPobiega:
Wrong base

Issue #198, opened 2016-06-08
Do not use transport connection for persistence if persistence is configured with different connection string


SzymonPobiega:
### Background

In V5 we used a dictionary keyed by connection string to get put/get the transport connection. Because of that, if SQL Server transport was configured to use different connection string than NH persistence, there was no sharing of connection.

### Problem

In V6 we use the `TransportTransaction` abstraction so currently if there is a transport SQL connection, NH persistence will use it even if it was configured for a different database.

### Solution

A solution to this problem is early closing of the transport connection. 

Suppose we close the connection before invoking the pipeline.
 * NHibernate persistence will not see the connection and create its own. If it has the same connection string as SQL transport, there will be no DTC escalation as everything happens within same `TransactionScope`
 * SQL server dispatcher will notice it has `TransactionScope` from the transport so it will create a new connection within that same scope when dispatching messages.

Everything should be good and the code should get much simpler. 

### Plan of attack

 * [x] SQL transport https://github.com/Particular/NServiceBus.SqlServer/pull/252
 * [x] NH Persistence https://github.com/Particular/NServiceBus.NHibernate/pull/200

SzymonPobiega:
A potential solution to this problem is early closing of the connection. 

Suppose we close the connection before invoking the pipeline.
 * NHibernate persistence will not see the connection and create its own. If it has the same connection string as SQL transport, there will be no DTC escalation as everything happens within same `TransactionScope`
 * SQL server dispatcher will notice it has `TransactionScope` from the transport so it will create a new connection within that same scope when dispatching messages.

Everything should be good and the code should get much simpler.

SzymonPobiega:
@Particular/nhibernate-persistence-maintainers @Particular/sqlserver-transport-maintainers please chime in.

SzymonPobiega:
@Particular/nhibernate-persistence-maintainers can we do it as maintainer budget without forming a proper task force?

SzymonPobiega:
I treat silence as a sign of acceptance.

PR #200, opened 2016-06-08
Removed the code that was responsible for sharing the connection instance

SzymonPobiega:
Removed the core responsible for sharing an instance of `SqlConnection`. Based on the changes done in https://github.com/Particular/NServiceBus.SqlServer/pull/252 the same goal will be achieved by having separate instances of `SqlConnection` in transport and in persistence provided that both use the same connection string. 

The key feature is early closing of transport connection that prevents DTC escalation when persistence connection is created.

### DIFF
--- a/src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateSynchronizedStorageAdapter.cs
+++ b/src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateSynchronizedStorageAdapter.cs
@@ -1,7 +1,7 @@
 ﻿namespace NServiceBus.Persistence.NHibernate
 {
     using System;
-    using System.Data.SqlClient;
+    using System.Data.Common;
     using System.Threading.Tasks;
     using global::NHibernate;
     using global::NHibernate.Impl;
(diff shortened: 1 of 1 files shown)


SzymonPobiega:
@Particular/nhibernate-persistence-maintainers can you review? I'll rebased in the meantime. The corresponding change in the SQL transport has already been merged.

SzymonPobiega:
This is now green on the build server (including ATs)

HEskandari:
Is this a change in behavior in v7? Should we mention it in the upgrade docs? The code LGTM. I'm merging it since the build is now green.

SzymonPobiega:
@HEskandari Maybe add something like this https://github.com/Particular/docs.particular.net/pull/1594 to the NH upgrade guide? Because we never know if people approach it from NH or SQL perspective.

SzymonPobiega:
Scratch that. On the NH side the behavior is the same as in V6: if in ambient transaction mode and using same connection string as SQL transport, no DTC escalation.

HEskandari:
@SzymonPobiega  :+1:

HEskandari:
With this merged, think we can close #198 and do another release?

PR #204, opened 2016-07-06
Aligning OutboxRecord with TransportOperations to save allocations

danielmarbach:
Relates to https://github.com/Particular/NServiceBus/pull/3903

Can be merged when core changes are merged

### DIFF
--- a/src/NServiceBus.NHibernate/Outbox/OutboxPersister.cs
+++ b/src/NServiceBus.NHibernate/Outbox/OutboxPersister.cs
@@ -1,7 +1,6 @@
 ﻿namespace NServiceBus.Persistence.NHibernate
 {
     using System;
-    using System.Collections.Generic;
     using System.Linq;
     using System.Threading.Tasks;
     using System.Transactions;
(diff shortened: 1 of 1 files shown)


danielmarbach:
@MarcinHoppe since you are on a spree. Could you rebase this on top of your changes in #206 ?

MarcinHoppe:
Will do.

MarcinHoppe:
@danielmarbach Rebased.

@Particular/nhibernate-persistence-maintainers This seems to be a minor change so I will take it out of WIP.

DavidBoike:
:+1:

danielmarbach:
What's up with this one?

SzymonPobiega:
@danielmarbach I am not convinced that in context of a NHibernate database call changing LINQ to for loops makes sense from performance point of view. Although I do understand we want to be as fast as possible in the core because core works all in-mem, here we have a cross-machine call (probably not only one). I prefer simplicity of LINQ (unless there is a way to prove the change gives us measurable benefit in terms of allocations.

Issue #209, opened 2016-08-11
Incorrect container lifecyle specified when using RegisterManagedSessionInTheContainer


WilliamBZA:
## Who's affected

Any V6 (core v5) user using `RegisterManagedSessionInTheContainer` in combination with Castle Windsor.

## Symptoms

When using `RegisterManagedSessionInTheContainer`, a `Func` is registered which returns the current NHibernate `ISession` allowing handlers to access the current ISession via the container.

This func is currently registered as `InstancePerCall`, implying that a new session is returned on each invocation.

However, the `NHibernateStorageContext` instance is the class providing the session, and this instance is registered as `InstancePerUnitOfWork`.

This mismatch causes occasional problems with Castle Windsor, as it tracks the transient (Instance Per Call) objects and throws an exception if the same instance of a transient type is returned twice.

## Proposed resolution

Change the lifecycle of the `ISession` func to match that of the underlying `NHibernateStorageContext` object. I.e. make the lifecyle `InstancePerUnitOfWork`

PR #210, opened 2016-08-11
Change the ISession lifecycle to instance per UOW

WilliamBZA:
Resolves https://github.com/Particular/NServiceBus.NHibernate/issues/209

:-1: to backporting. This is applicable to a very specific class of user, and there is a workaround available too (manually register the ISession in the container).

Ping @Particular/nhibernate-persistence-maintainers


SzymonPobiega:
@WilliamBZA looks good. I've realized there was no AT for that whole API. I've added a test. Can you verify if the test is actually doing similar thing as a user would do? If so, I'll merge.

WilliamBZA:
The test looks good, but doesn't seem related to the original issue :smile: 

Reading the test, it validates that a manually configured NHibernate Cfg can be registered via `RegisterManagedSessionInTheContainer`, and injected via property injection?

SzymonPobiega:
@WilliamBZA the thing I wanted to validate is that the `ISession` lifecycle is OK. I send two messages to validate that the session injected to the second handler will be the brand new one, not the one used previously (and closed).

WilliamBZA:
Ping @SzymonPobiega or @Particular/nhibernate-persistence-maintainers :smile:

SzymonPobiega:
@WilliamBZA sorry for the delay :(

Issue #207, opened 2016-08-03
Provide code-first API for setting outbox data retention period


SzymonPobiega:
Similar RavenDB issue: https://github.com/Particular/NServiceBus.RavenDB/issues/262

SzymonPobiega:
Duplicate of #248

PR #219, opened 2016-09-02
Make Outbox settings available as extension methods on .EnableOutbox()

DavidBoike:
Connects to #207

Related to https://github.com/Particular/NServiceBus.RavenDB/pull/271 and using the same API.

This would give a code option in addition to the current appSetting configuration option.

@Particular/nhibernate-persistence-maintainers @Particular/ravendb-persistence-maintainers

### DIFF
--- a/src/NServiceBus.NHibernate/Outbox/NHibernateOutboxExtensions.cs
+++ b/src/NServiceBus.NHibernate/Outbox/NHibernateOutboxExtensions.cs
@@ -1,0 +1,43 @@
+﻿namespace NServiceBus.NHibernate.Outbox
+{
+    using System;
+    using NServiceBus.Configuration.AdvanceExtensibility;
+    using NServiceBus.Outbox;
+
+    /// <summary>
+    /// Contains extensions methods which allow to configure RavenDB outbox specific configuration
+    /// </summary>
+    public static class NHibernateOutboxExtensions
+    {
+        /// <summary>
+        /// Sets the time to keep the deduplication data to the specified time span.
+        /// </summary>
+        /// <param name="configuration">The configuration being extended</param>
+        /// <param name="timeToKeepDeduplicationData">The time to keep the deduplication data. 
+        /// The cleanup process removes entries older than the specified time to keep deduplication data, therefore the time span cannot be negative</param>
+        /// <returns>The configuration</returns>
+        public static OutboxSettings SetTimeToKeepDeduplicationData(this OutboxSettings configuration, TimeSpan timeToKeepDeduplicationData)
+        {
+            var now = DateTime.UtcNow;
+            if (now - timeToKeepDeduplicationData >= now)
+            {
+                throw new ArgumentException("Specify a non-negative TimeSpan. The cleanup process removes entries older than the specified time to keep deduplication data, therefore the time span cannot be negative.", "timeToKeepDeduplicationData");
+            }
+
+            configuration.GetSettings().Set("Outbox.TimeToKeepDeduplicationData", timeToKeepDeduplicationData);
+            return configuration;
+        }
+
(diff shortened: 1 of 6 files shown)


src/NServiceBus.NHibernate/Outbox/NHibernateOutboxExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// Sets the time to keep the deduplication data to the specified time span.
+        /// </summary>
+        /// <param name="configuration">The configuration being extended</param>
+        /// <param name="timeToKeepDeduplicationData">The time to keep the deduplication data. 
+        /// The cleanup process removes entries older than the specified time to keep deduplication data, therefore the time span cannot be negative</param>
+        /// <returns>The configuration</returns>
+        public static OutboxSettings SetTimeToKeepDeduplicationData(this OutboxSettings configuration, TimeSpan timeToKeepDeduplicationData)
+        {
+            var now = DateTime.UtcNow;
+            if (now - timeToKeepDeduplicationData >= now)
```
SzymonPobiega: Would `timeToKeep < TimeSpan.Zero` work?

src/NServiceBus.NHibernate/Outbox/OutboxCleaner.cs
```diff
@@ -21,7 +23,7 @@ protected override Task OnStart(IMessageSession busSession)
 
             if (configValue == null)
             {
-                timeToKeepDeduplicationData = TimeSpan.FromDays(7);
+                timeToKeepDeduplicationData = settings.GetOrDefault<TimeSpan?>("Outbox.TimeToKeepDeduplicationData") ?? TimeSpan.FromDays(7);
```
SzymonPobiega: I think in most similar classes we tend to put the setting access code in the feature and pass the actual value to the startup task. It would keep the FST cleaner.

SzymonPobiega:
@DavidBoike ping. I think this is the last thing we want in the RC.

src/NServiceBus.NHibernate/Outbox/NHibernateOutboxExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+            return configuration;
+        }
+
+        /// <summary>
+        /// Sets the frequency to run the deduplication data cleanup task.
+        /// </summary>
+        /// <param name="configuration">The configuration being extended</param>
+        /// <param name="frequencyToRunDeduplicationDataCleanup">The frequency to run the deduplication data cleanup task. By specifying a negative time span (-1) the cleanup task will never run.</param>
+        /// <returns>The configuration</returns>
+        public static OutboxSettings SetFrequencyToRunDeduplicationDataCleanup(this OutboxSettings configuration, TimeSpan frequencyToRunDeduplicationDataCleanup)
```
andreasohlund: Isn't "set" redundant?

How about just `.DeduplicationDataCleanupFrequency(Timespan.FromHours(1))` ?

src/NServiceBus.NHibernate/Outbox/OutboxCleaner.cs
andreasohlund: Agreed, in general I believe the idea it to "Not use settings in the runtime code". Ie in the features translate settings into explicit args passed to pipeline stuff, startuptasks etc

DavidBoike:
@SzymonPobiega @andreasohlund addressed your comments.

Going to remove the word `Set` from the API in Raven as well to match.

DavidBoike:
Because of difficulties in API coordination between the Core and RavenDB, there's no good way to structure this API given that we're so close to RC. We'll have to live with it for now.

PR #220, opened 2016-09-05
Initialize the subscription storage

danielmarbach:
The subscription storage has an Init method which is never called. The core allows implementing `IInitializableSubscriptionStorage`. Then the core calls init on that type. I'm binding the class directly to the container. In theory with the interface scanning it should work. 

Happy to change it to the interface if you prefer.

@Particular/nhibernate-persistence-maintainers please review


SzymonPobiega:
Good catch @danielmarbach . But wonder if we should depend on the core here. We can create a small FST that would call it. I have a feeling this `Initializable` thing will be removed soon anyway.

danielmarbach:
Good for me. You'll drive it then?

SzymonPobiega:
@danielmarbach if you have a moment, can you check my changes?

danielmarbach:
Looks good!

Issue #201, opened 2016-06-08
Using SQL Azure for persistence fails for V6 with NRE


ramonsmits:
If using SQL Azure this fails with the following NullReferenceException with NServiceBus V6. Not an issue with V5.

It looks like it is related to promoting to a distributed transaction. The transport used is MSMQ. 
# Exception info

```
2016-06-08 13:13:50.9464|INFO|12|NServiceBus.FirstLevelRetriesBehavior|First Level Retry is going to retry message '58fcada3-b163-4f12-a4ca-a61f00fafc62' because of an exception:
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Transactions.Transaction.GetPromotedToken()
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Data.SqlClient.SqlInternalConnection.EnlistNonNull(Transaction tx)
   at System.Data.SqlClient.SqlInternalConnection.Enlist(Transaction tx)
   at System.Data.ProviderBase.DbConnectionInternal.ActivateConnection(Transaction transaction)
   at System.Data.ProviderBase.DbConnectionPool.PrepareConnection(DbConnection owningObject, DbConnectionInternal obj, Transaction transaction)
   at System.Data.ProviderBase.DbConnectionPool.TryGetConnection(DbConnection owningObject, UInt32 waitForMultipleObjectsTimeout, Boolean allowCreate, Boolean onlyOneCheckConnection, DbConnectionOptions userOptions, DbConnectionInternal& connection)
   at System.Data.ProviderBase.DbConnectionPool.TryGetConnection(DbConnection owningObject, TaskCompletionSource`1 retry, DbConnectionOptions userOptions, DbConnectionInternal& connection)
[...]

PR #221, opened 2016-09-07
Made synchronized storage session lazy

SzymonPobiega:
@ramonsmits @Particular/nhibernate-persistence-maintainers please review. This makes the synchronized storage session lazy (similar to V5 behavior).

### DIFF
--- a/src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateAmbientTransactionSynchronizedStorageSession.cs
+++ b/src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateAmbientTransactionSynchronizedStorageSession.cs
@@ -8,40 +8,30 @@
     using NServiceBus.Persistence;
 
     [SkipWeaving]
-    class NHibernateAmbientTransactionSynchronizedStorageSession : CompletableSynchronizedStorageSession, INHibernateSynchronizedStorageSession
+    class NHibernateLazyAmbientTransactionSynchronizedStorageSession : CompletableSynchronizedStorageSession, INHibernateSynchronizedStorageSession
     {
-        Func<IDbConnection> connectionFactory;
-        Func<IDbConnection, ISession> sessionFactory;
-        ISession session;
-        IDbConnection connection;
+        Lazy<ISession> session;
+        Lazy<IDbConnection> connection;
 
-        public NHibernateAmbientTransactionSynchronizedStorageSession(Func<IDbConnection> connectionFactory, Func<IDbConnection, ISession> sessionFactory)
+        public NHibernateLazyAmbientTransactionSynchronizedStorageSession(Func<IDbConnection> connectionFactory, Func<IDbConnection, ISession> sessionFactory)
         {
-            this.connectionFactory = connectionFactory;
-            this.sessionFactory = sessionFactory;
+            connection = new Lazy<IDbConnection>(connectionFactory);
+            session = new Lazy<ISession>(() => sessionFactory(connection.Value));
         }
 
-        public ISession Session
-        {
-            get
-            {
-                if (connection == null)
-                {
-                    connection = connectionFactory();
-                }
-                if (session == null)
-                {
-                    session = sessionFactory(connection);
-                }
-                return session;
-            }
-        }
+        public ISession Session => session.Value;
 
         public void Dispose()
         {
-            session?.Flush();
-            session?.Dispose();
-            connection?.Dispose();
+            if (session.IsValueCreated)
+            {
+                session.Value.Flush();
+                session.Value.Dispose();
+            }
+            if (connection.IsValueCreated)
+            {
+                connection.Value.Dispose();
+            }
         }
 
         public Task CompleteAsync()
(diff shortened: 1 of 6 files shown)


SzymonPobiega:
The only downside I can see is that it can blow up in a handler (rather then in the open session behavior). But I guess this is a fair compromise as it saves establishing a connection if user is not using storage session in the handler.

ramonsmits:
@SzymonPobiega I tested this PR and published the results here https://docs.google.com/spreadsheets/d/1OjK4QVMClj3CoYJFKqu6tI2eDwhVnhVzUjI9BEn4bws/edit#gid=0

Summary: Performance is pretty equal to the NHibernate configuration hack with just one test run.

ramonsmits:
This also fixes  #201

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateAmbientTransactionSynchronizedStorageSession.cs
```diff
@@ -9,28 +10,40 @@
...
+        public ISession Session
         {
-            Session.Flush();
-            Session.Dispose();
-
-            if (ownsConnection)
+            get
             {
-                connection.Dispose();
+                if (connection == null)
```
ramonsmits: Why two if statements? Aren't these both null or non null?

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateLazyNativeTransactionSynchronizedStorageSession.cs
```diff
@@ -0,0 +1,47 @@
...
+        public NHibernateLazyNativeTransactionSynchronizedStorageSession(Func<ISession> sessionFactory)
+        {
+            this.sessionFactory = sessionFactory;
+        }
+
+        public ISession Session
+        {
+            get
+            {
+                if (session == null)
```
ramonsmits: Why not make use of Lazy<T> to reveal the intent?

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateLazyNativeTransactionSynchronizedStorageSession.cs
```diff
@@ -0,0 +1,47 @@
...
+                }
+                return session;
+            }
+        }
+
+        public ITransaction Transaction => Session.Transaction;
+
+        public Task CompleteAsync()
+        {
+            transaction?.Commit();
```
ramonsmits: Use Laze<T>.IsValueCreated and if so assume these values not be null?

ramonsmits:
Besides the non-critical notes it LGTM.

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateAmbientTransactionSynchronizedStorageSession.cs
SzymonPobiega: In theory creating a connection and opening a session are two operations and the failure might happen between these calls. I guess in normal scenario the flow would be interrupted anyway but if somebody does:

```
while (!success)
{
try
{
   var s = session.Session;
   success = true;
}
catch (Exception) {}
}
```

then this code will behave better. Hmm.. after writing this I think that would be a very unusual scenario. I'll collapse these ifs.

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateLazyNativeTransactionSynchronizedStorageSession.cs
SzymonPobiega: `Lazy` is thread safe but we don't need thread safety here. Do you still think it would be read better? If so, I can change it. No strong opinions.

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateAmbientTransactionSynchronizedStorageSession.cs
ramonsmits: If it's an issue another approach is the not directly assign them but store them in scope first and if both succeed to then copy those references.

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateLazyNativeTransactionSynchronizedStorageSession.cs
ramonsmits: I personally prefer this for readability:

```
public ISession Session
{
    get { return lazySession.Value; }
}
```

We don't need the thread safety indeed but that overhead can be prevented by using the correct overload.

SzymonPobiega:
@Particular/nhibernate-persistence-maintainers what do you think about this?

MarcinHoppe:
@SzymonPobiega It looks good to me. I agree with @ramonsmits that using `Lazy` would make the code easier to understand (relation between class name and implementation would be easier to see and analyze).

SzymonPobiega:
@MarcinHoppe @ramonsmits changed it to proper lazy `Lazy`. Please take a look.

PR #225, opened 2016-09-27
Make timeout manager next chunk slice calculation consistent with other persisters

### LINKED ISSUE #145: TimeoutPersister should use '>' rather than '>=' when applying slice predicate
Currently the `TimeoutPersister` uses the following predicate to filter out timeouts:
```
 .And(x => x.Time >= startSlice && x.Time <= now)
```
https://github.com/Particular/NServiceBus.NHibernate/blob/develop/src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutPersister.cs#L40

`ExpiredTimeoutsPoller` issues queries against `IQueryTimeouts` with the oldest of timeouts returned in the previous query. Having that logic combined with the predicate above, results in possibility of duplicates, when two succeeding queries are issued before the timeout entity is deleted. For instance:
* EpiredTimeoutsPoller queries with the initial startSlice 
* TimeoutPersister returns timeouts with times d1, d2, d3
* EpiredTimeoutsPoller saves d3 as the next startSlice, later on queries again
* TimeoutPersister is queried with d3, returning the timeout with d3 again.

That predicate is different from one in 'InMemoryTimeoutPersister' which uses '>', see: https://github.com/Particular/NServiceBus/blob/48cad8bbd3c2e0fff4ec6b6abd084c0b78b7bbd8/src/NServiceBus.Core/Persistence/InMemory/TimeoutPersister/InMemoryTimeoutPersister.cs#L110

MarcinHoppe:
Fixes #145

### DIFF
--- a/src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutPersister.cs
+++ b/src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutPersister.cs
@@ -41,7 +41,7 @@ namespace NServiceBus.TimeoutPersisters.NHibernate
                 {
                     var results = session.QueryOver<TimeoutEntity>()
                         .Where(x => x.Endpoint == EndpointName)
-                        .And(x => x.Time >= startSlice && x.Time <= now)
+                        .And(x => x.Time > startSlice && x.Time <= now)
                         .OrderBy(x => x.Time).Asc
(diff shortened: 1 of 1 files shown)


MarcinHoppe:
@Particular/nhibernate-persistence-maintainers Please review.

PR #233, opened 2016-10-17
[WIP] Add retries to subscription storage operation

### LINKED ISSUE #232: Gracefully handle concurrency errors with scaled-out subscribers
Scaled-out subscribers may cause concurrency errors (PK violation) during processing of subscription messages by the publisher.

These errors should be gracefully handled within the persister, as it is currently done RavenDB and ASP persisters:

https://github.com/Particular/NServiceBus.Persistence.AzureStorage/blob/develop/src/NServiceBus.Persistence.AzureStorage/Subscriptions/AzureSubscriptionStorage.cs#L62

https://github.com/Particular/NServiceBus.RavenDB/blob/develop/src/NServiceBus.RavenDB/Subscriptions/SubscriptionPersister.cs#L31

The faulty tests have all been moved to the `Acceptance tests - quarantine build configuration`.

MarcinHoppe:
Fixes #232

### DIFF
--- a/src/NServiceBus.NHibernate/Subscriptions/SubscriptionPersister.cs
+++ b/src/NServiceBus.NHibernate/Subscriptions/SubscriptionPersister.cs
@@ -7,12 +7,10 @@ namespace NServiceBus.Unicast.Subscriptions.NHibernate
     using System.Transactions;
     using global::NHibernate;
     using global::NHibernate.Criterion;
-    using global::NHibernate.Exceptions;
     using Logging;
     using MessageDrivenSubscriptions;
     using NServiceBus.Extensibility;
     using IsolationLevel = System.Data.IsolationLevel;
-    using TransactionException = global::NHibernate.TransactionException;
 
     class SubscriptionPersister : ISubscriptionStorage
     {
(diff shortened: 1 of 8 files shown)


MarcinHoppe:
@Particular/nhibernate-persistence-maintainers Please review and merge.

src/NServiceBus.NHibernate/Subscriptions/SubscriptionPersister.cs
```diff
@@ -25,6 +27,37 @@ public SubscriptionPersister(ISessionFactory sessionFactory)
...
+                    }
+                    else
+                    {
+                        // This was the last attempt, give up.
+                        throw;
+                    }
+                }
+            }
+
+            throw new Exception($"Internal error occured when storing subscription of endpoint '{subscriber.Endpoint}' to message '{messageType}'.");
```
SzymonPobiega: When do we get here?

MarcinHoppe: We should never get here. If we do get here, it's a bug.

SzymonPobiega:
I think (but I am not 100% sure) we need to add similar thing for the unsubscribe because `Delete` in NH expects 1 row affected. In the scale-out scenario we might have multiple attempts to delete the same row, only one of which will succeed.

MarcinHoppe:
@SzymonPobiega Doesn't [this loop](http://infoshareacademy.com/zostan_trenerem.html) delete `Subscription` records one-by-one?

SzymonPobiega:
@MarcinHoppe still working on this?

MarcinHoppe:
@SzymonPobiega Yes, I'm still on it. ATTs fail regularly on TeamCity (but Work on My Machine (TM)) and I am trying to investigate why that is.

Issue #241, opened 2016-11-15
OutboxCleaner.PerformCleanup tears down the endpoint


timbussmann:
It seems that a connection failure during `OutboxCleaner.PerformCleanup` will tear down the endpoint, as exceptions aren't catched and will bubble up to the executing thread. See the issue here: https://github.com/Particular/NServiceBus/issues/4293

Given the fact, that we use circuit breakers to handle persistence failures in other places (e.g. the timeout manager) before directly shutting down the endpoint, I think it would be reasonable to do the same with the cleanup task. It probably makes sense to still fail the endpoint after the cleanup task wasn't able to run for a specified amount of time (several hours?).

ping @Particular/nhibernate-persistence-maintainers 

### Plan of Attack

- [ ] Add a circuit breaker to the Outbox cleanup task: #242 
- [x] Decide on how to configure the time to wait before triggering the circuit breaker
- [ ] Update the doco
  - [ ] [Critical exception wait time](https://docs.particular.net/nservicebus/recoverability/critical-exception-for-timeout-outages)
  - [ ] [Outbox with NHibernate persistence](https://docs.particular.net/nservicebus/nhibernate/outbox)

MarcinHoppe:
@timbussmann @ramonsmits @andreasohlund I agree this is a good change for 7.1 (added this issue to the milestone). Is there anything that makes this issue urgent?

timbussmann:
I'll leave that classification to you. As far as I understand it, it can cause the endpoint to shut down earlier than any configured circuit breaker (e.g. is the SQL server is up again after 1.5 minutes, the timeout managers circuit breaker is not violated and the endpoint would recover, while the outbox cleanup task will tear down then endpoint). So theoretically this could be seen as a critical, but in most cases, fixing this would just postpone the endpoint shutdown by ~1 minute.

andreasohlund:
I'd lean towards a patch for the reasons mentioned by Tim

MarcinHoppe:
Thanks.

@HEskandari @SzymonPobiega What do you think about it?

HEskandari:
I'm 👍 on the patch. If it crashes the endpoint, should be considered as critical IMO.

MarcinHoppe:
We will still crash the endpoint, only later and not that often. I'm also 👍 on a patch.

MarcinHoppe:
I created a mileston for the patch and assigned this issue to this new milestone.

MarcinHoppe:
[...]

PR #242, opened 2016-11-15
Outbox cleanup circuit breaker

ramonsmits:
Quick draft to resolve https://github.com/Particular/NServiceBus.NHibernate/issues/241

I have put the circuit breaker creation in the cleaner as that already contains configuration. In core circuitbreakers are created in the feature which seems to make more sense but that felt strange to do here unless that configuration value retrieval code is moved to the feature too.

### DIFF
--- a/src/NServiceBus.NHibernate/Outbox/OutboxCleaner.cs
+++ b/src/NServiceBus.NHibernate/Outbox/OutboxCleaner.cs
@@ -7,9 +7,11 @@ namespace NServiceBus.Features
 
     class OutboxCleaner : FeatureStartupTask
     {
-        public OutboxCleaner(INHibernateOutboxStorage outboxPersister)
+
+        public OutboxCleaner(INHibernateOutboxStorage outboxPersister, CriticalError criticalError)
         {
             this.outboxPersister = outboxPersister;
+            this.criticalError = criticalError;
         }
 
         protected override Task OnStart(IMessageSession busSession)
@@ -36,6 +38,25 @@ namespace NServiceBus.Features
                 throw new Exception("Invalid value in \"NServiceBus/Outbox/NHibernate/FrequencyToRunDeduplicationDataCleanup\" AppSetting. Please ensure it is a TimeSpan.");
             }
 
+            var key = "NServiceBus/Outbox/NHibernate/TimeToWaitBeforeTriggeringCriticalErrorWhenCleanupTaskFails";
+
+            configValue = ConfigurationManager.AppSettings.Get(key);
+
+            if (configValue == null)
+            {
+                timeToWaitBeforeTriggeringCriticalError = TimeSpan.FromMinutes(2);
+            }
+            else if (!TimeSpan.TryParse(configValue, out timeToWaitBeforeTriggeringCriticalError))
+            {
+                throw new Exception($"Invalid value in \"{key}\" AppSetting. Please ensure it is a TimeSpan.");
+            }
+
+            circuitBreaker = new RepeatedFailuresOverTimeCircuitBreaker(
+                "OutboxCleanupTaskConnectivity",
+                timeToWaitBeforeTriggeringCriticalError,
+                ex => criticalError.Raise("Repeated failures when purging expired outbox records from storage, endpoint will be terminated.", ex)
+                );
+
             cleanupTimer = new Timer(PerformCleanup, null, TimeSpan.FromMinutes(1), frequencyToRunDeduplicationDataCleanup);
 
             return Task.FromResult(true);
@@ -55,7 +76,15 @@ namespace NServiceBus.Features
 
         void PerformCleanup(object state)
         {
-            outboxPersister.RemoveEntriesOlderThan(DateTime.UtcNow - timeToKeepDeduplicationData);
+            try
+            {
+                outboxPersister.RemoveEntriesOlderThan(DateTime.UtcNow - timeToKeepDeduplicationData);
+                circuitBreaker.Success();
+            }
+            catch (Exception ex)
+            {
+                circuitBreaker.Failure(ex);
+            }
         }
 
         // ReSharper disable NotAccessedField.Local
@@ -64,5 +93,8 @@ namespace NServiceBus.Features
         TimeSpan frequencyToRunDeduplicationDataCleanup;
         INHibernateOutboxStorage outboxPersister;
         TimeSpan timeToKeepDeduplicationData;
+        RepeatedFailuresOverTimeCircuitBreaker circuitBreaker;
+        TimeSpan timeToWaitBeforeTriggeringCriticalError;
+        CriticalError criticalError;
     }
 }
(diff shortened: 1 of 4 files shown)


MarcinHoppe:
I rebased this PR on top of develop.

MarcinHoppe:
@ramonsmits @Particular/nhibernate-persistence-maintainers Can we use the value that is set using `TimeToWaitBeforeTriggeringCriticalErrorOnTimeoutOutages`? See doco here:

https://docs.particular.net/nservicebus/recoverability/critical-exception-for-timeout-outages

It also applies to the persister, so it would make sense to have the circuit breaker behave in the same way for the Timeouts and well as the Outbox.

MarcinHoppe:
We could rename the method to not include the Timeouts part in core 6.1 (obsolete with a warning).

MarcinHoppe:
@ramonsmits @Particular/nhibernate-persistence-maintainers I think this PR is good to go, assuming that we want to configure the circuit breaker via AppSettings.

Please chime in [here](https://github.com/Particular/NServiceBus.NHibernate/pull/242#issuecomment-262754254).

ramonsmits:
>  Can we use the value that is set using TimeToWaitBeforeTriggeringCriticalErrorOnTimeoutOutages

@MarcinHoppe I intentionally didn't as its two independent features that would then share implicit schema. Also, the timeout storage is only applicable if the transport doesn't support timeouts natively.

MarcinHoppe:
Also, timeouts and Oubox can use two different persisters, so it wouldn't always be good to share the circuit breaker timeout. I think the current API (AppConfig) is good enough and consistent with other Outbox configuration parameters.

MarcinHoppe:
@SzymonPobiega @HEskandari Can you give this PR a look and merge?

src/NServiceBus.NHibernate/Outbox/OutboxCleaner.cs
```diff
@@ -36,6 +37,25 @@ protected override Task OnStart(IMessageSession busSession)
                 throw new Exception("Invalid value in \"NServiceBus/Outbox/NHibernate/FrequencyToRunDeduplicationDataCleanup\" AppSetting. Please ensure it is a TimeSpan.");
             }
 
+            var key = "NServiceBus/Outbox/NHibernate/TimeToWaitBeforeTriggeringCriticalErrorWhenCleanupTaskFails";
```
SzymonPobiega: `const`?

src/NServiceBus.NHibernate/Outbox/OutboxCleaner.cs
```diff
@@ -36,6 +37,25 @@ protected override Task OnStart(IMessageSession busSession)
...
+
+            configValue = ConfigurationManager.AppSettings.Get(key);
+
+            if (configValue == null)
+            {
+                timeToWaitBeforeTriggeringCriticalError = TimeSpan.FromMinutes(2);
```
SzymonPobiega: Extract `TimeSpan.FromMinutes(2)` as a readonly field to give it more visibility?

SzymonPobiega:
Few minor things. Otherwise LGTM.

SzymonPobiega:
@ramonsmits @MarcinHoppe ping.

src/NServiceBus.NHibernate/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -0,0 +1,68 @@
...
+        public void Success()
+        {
+            var oldValue = Interlocked.Exchange(ref failureCount, 0);
+
+            if (oldValue == 0)
+            {
+                return;
+            }
+
+            timer.Change(System.Threading.Timeout.Infinite, System.Threading.Timeout.Infinite);
```
HEskandari: This is a bit hard to read and capture the intent. Would it be a good idea to have a properly named/intent capturing private method? This is subjective of course so feel free to ignore if you disagree.

src/NServiceBus.NHibernate/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        public void Failure(Exception exception)
+        {
+            lastException = exception;
+            var newValue = Interlocked.Increment(ref failureCount);
+
+            if (newValue == 1)
+            {
+                timer.Change(timeToWaitBeforeTriggering, NoPeriodicTriggering);
```
HEskandari: Same as the above.

src/NServiceBus.NHibernate/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        long failureCount;
+        Exception lastException;
+
+        string name;
+        Timer timer;
+        TimeSpan timeToWaitBeforeTriggering;
+        Action<Exception> triggerAction;
+
+        static TimeSpan NoPeriodicTriggering = TimeSpan.FromMilliseconds(-1);
```
HEskandari: Isn't this the same as Timeout.Infinite?

src/NServiceBus.NHibernate/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            if (Interlocked.Read(ref failureCount) > 0)
+            {
+                Logger.WarnFormat("The circuit breaker for {0} will now be triggered", name);
+                triggerAction(lastException);
+            }
+        }
+
+        long failureCount;
+        Exception lastException;
```
HEskandari: Is the last exception enough? or should we go with an AggregateException instead? That'd require a bit more code to ensure exceptions are aggregated correctly and cleared when successful.

src/NServiceBus.NHibernate/Outbox/OutboxCleaner.cs
MarcinHoppe: I actually inlined this variable to be consistent with how other configuration keys are handled. Key values seem to be descriptive and only used locally in this file. Since they don't change often (doco), encapsulating them brings little value.

Agree?

src/NServiceBus.NHibernate/Outbox/OutboxCleaner.cs
MarcinHoppe: Extracted this and all the other `TimeSpans` in this file into separate `static readonly` fields.

MarcinHoppe:
@HEskandari The circuit breaker code has been copied verbatim from the core. Do we want to impove upon the code and risk these two copies getting out of sync?

I'd rather do a PR in the core and then copy what they come up with. Thoughts?

HEskandari:
@MarcinHoppe fair enough.

MarcinHoppe:
@HEskandari OK. Can I dismiss your request for changes?

@SzymonPobiega If you are OK with my latest changes, this is ready for merge (build is green).

SzymonPobiega:
@MarcinHoppe done.

PR #253, opened 2017-01-24
Fixes the schema generation bug

SzymonPobiega:
* Uses `OptimizedSchemaUpdate` for all schema updates
 * Adds condition to `OptimizedSchemaUpdate` to only modify timeout table PK

### DIFF
--- a/src/NServiceBus.NHibernate.Tests/Outbox/When_adding_outbox_messages.cs
+++ b/src/NServiceBus.NHibernate.Tests/Outbox/When_adding_outbox_messages.cs
@@ -9,12 +9,10 @@ namespace NServiceBus.NHibernate.Tests.Outbox
     using global::NHibernate.Cfg;
     using global::NHibernate.Mapping.ByCode;
     using global::NHibernate.Mapping.ByCode.Conformist;
-    using global::NHibernate.Tool.hbm2ddl;
     using NServiceBus.Extensibility;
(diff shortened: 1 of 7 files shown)


SzymonPobiega:
@Particular/nhibernate-persistence-maintainers please review. I decided to go with no tests because I could not figure out a decent way to test it, other than running a bunch of integration-level scenarios, each after full DB clleanup and these would be too costly IMO.

MarcinHoppe:
@SzymonPobiega: this PR targets `master`. Was this the original intent or was it `develop`?

SzymonPobiega:
@MarcinHoppe I was not sure when I created it, if this should be a patch or not. What do you think?

MarcinHoppe:
@SzymonPobiega I'd say develop + patch the version that customer has (6.2.7 IIRC).

SzymonPobiega:
This solution won't work in case of non-default schema. In that case the SQL instructions contain also the schema name and `Replace` won't work. The alternative would be to use clever regular expressions.

MarcinHoppe:
@SzymonPobiega This can be closed, it will be superseded by #270.

SzymonPobiega:
@MarcinHoppe OK. Let me know when you need a second pair of 👀

MarcinHoppe:
7.1.4, 7.0.6, and 6.2.8 have been released to address this. Closing.

Issue #255, opened 2017-02-01
Database connection is not closed when synchronized storage session flush throws


MarcinHoppe:
## Symptoms

When [business data access with NHibernate persistence](https://docs.particular.net/nservicebus/nhibernate/accessing-data) fails, the database connection opened for the failed NHibernate session will not be closed.

Consistent rate of data access errors leads to ADO .NET connection pool exhaustion and endpoint shutdown.

## Who is affected?

Users of NHibernate persistence that use `SynchronizedStorageSession` are affected.

## Plan of Attack

- [x] Release 7.0.4 hotfix PR #256 
- [x] Release 7.1.2 hotfix PR #257
- [x] Fix on develop
- [x] Announce

MarcinHoppe:
7.0.4 and 7.1.2 have been released. Closing.

PR #256, opened 2017-02-01
Close connection even if session flush throws

MarcinHoppe:
## Symptoms

When [business data access with NHibernate persistence](https://docs.particular.net/nservicebus/nhibernate/accessing-data) fails, the database connection opened for the failed NHibernate session will not be closed.

Consistent rate of data access errors leads to ADO .NET connection pool exhaustion and endpoint shutdown.

## Who is affected?

Users of NHibernate persistence that use `SynchronizedStorageSession` are affected.

Related to #255

### DIFF
--- a/src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateAmbientTransactionSynchronizedStorageSession.cs
+++ b/src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateAmbientTransactionSynchronizedStorageSession.cs
@@ -23,14 +23,20 @@
 
         public void Dispose()
         {
-            if (session.IsValueCreated)
+            try
             {
-                session.Value.Flush();
-                session.Value.Dispose();
+                if (session.IsValueCreated)
+                {
+                    session.Value.Flush();
+                    session.Value.Dispose();
+                }
             }
-            if (connection.IsValueCreated)
+            finally
             {
-                connection.Value.Dispose();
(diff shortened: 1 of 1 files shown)


src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateAmbientTransactionSynchronizedStorageSession.cs
```diff
@@ -23,14 +23,20 @@ public NHibernateLazyAmbientTransactionSynchronizedStorageSession(Func<IDbConnec
 
         public void Dispose()
         {
-            if (session.IsValueCreated)
+            try
```
SimonCropp: shouldn't a single IsValueCreated be outside the tryf? rather then two checks inside?

MarcinHoppe: Looks like connection will never be created if session has not been accessed, so check on session should be enough. You have a good eye @SimonCropp.

I'd leave these more conservative checks in hotfixes and go with a single `if` on develop. Thoughts?

SimonCropp: > I'd leave these more conservative checks in hotfixes and go with a single if on develop. Thoughts?

sounds good

PR #262, opened 2017-02-15
Use CombGuid as timeout identifier

### LINKED ISSUE #260: v7 for NSB core v6 doesn't use CombGuids for TimeoutEntity any more
I noticed that with the big change for NSB core v6 the method `TimeoutPersister.Add()` was changed from using the method `GenerateCombGuid()` to just use `Guid.NewGuid()`. Was this done on purpose?

I think this is wrong since now the primary key isn't an increasing value any more which could cause performance problems in MS SQL server.

Note: If you reintroduce the CombGuid, don't just add the method from before again. This method has the daylight saving time bug described in Particular/NServiceBus#4195. Either copy the fixed code from core or let NHibernate do the work by as shown in #113

MarcinHoppe:
Fixes #260 on develop

### DIFF
--- a/src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutPersister.cs
+++ b/src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutPersister.cs
@@ -86,23 +86,23 @@ namespace NServiceBus.TimeoutPersisters.NHibernate
 
         public async Task Add(TimeoutData timeout, ContextBag context)
         {
-            var timeoutId = Guid.NewGuid();
             using (var session = await OpenSession(context).ConfigureAwait(false))
             {
-                session.Session().Save(new TimeoutEntity
+                var timeoutEntity = new TimeoutEntity
                 {
-                    Id = timeoutId,
                     Destination = timeout.Destination,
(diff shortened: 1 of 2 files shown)


src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutPersister.cs
```diff
@@ -86,23 +86,23 @@ public Task<TimeoutsChunk> GetNextChunk(DateTime startSlice)
...
                     Destination = timeout.Destination,
                     SagaId = timeout.SagaId,
                     State = timeout.State,
                     Time = timeout.Time,
                     Headers = ConvertDictionaryToString(timeout.Headers),
                     Endpoint = timeout.OwningTimeoutManager
-                });
+                };
+
+                session.Session().Save(timeoutEntity);
```
SzymonPobiega: Save returns the ID. I think it would be move obvious to use this one instead of getting it again from the property on the entity.

MarcinHoppe: @SzymonPobiega Fixed.

PR #266, opened 2017-02-21
Added a hook for implementing UoW on top of SynchronizedStorageSession

SzymonPobiega:
Adds `RegisterCommitHook` method to the storage session. It allows registering custom unit of work completion logic that is executed before committing the storage session transaction.


src/NServiceBus.NHibernate/SynchronizedStorage/INHibernateSynchronizedStorageSession.cs
```diff
@@ -1,9 +1,13 @@
...
 {
+    using System;
+    using System.Threading.Tasks;
     using global::NHibernate;
 
     interface INHibernateSynchronizedStorageSession
     {
         ISession Session { get; }
+
+        void RegisterCommitHook(Func<Task> callback);
```
andreasohlund: Would there be value in trying to align with NH event listener lingo

http://nhibernate.info/doc/nhibernate-reference/events.html

In that case perhaps: `OnCommit(...)` ?

SzymonPobiega: @andreasohlund I don't think we should aim for alignment as the goal would be to have the same hook in native SQL persistence.  On the other hand, I am not attached to the current name so if you thing `OnCommit` is better, I am 👍 to change.

andreasohlund: Wouldn't we want  to call this `.OnSaveChanges(...)` in raven?

SzymonPobiega:
@Particular/nhibernate-persistence-maintainers please take a look. I think this is ready.

SzymonPobiega:
Ping :-) cc @tmasternak

src/NServiceBus.NHibernate/CallbackList.cs
```diff
@@ -0,0 +1,48 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    class CallbackList
+    {
+        Func<Task> singleItem;
+        List<Func<Task>> multipleItems;
+
+        public void Add(Func<Task> callback)
```
MarcinHoppe: Is the effect of storing single item separately measurable? It would be good if this complexity was justified by a significant benefit.

src/NServiceBus.NHibernate/CallbackList.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            if (singleItem != null)
+            {
+                await singleItem().ConfigureAwait(false);
+            }
+            else if (multipleItems != null)
+            {
+                foreach (var callback in multipleItems)
+                {
+                    await callback().ConfigureAwait(false);
```
MarcinHoppe: Is there a reason to run these callbacks asynchronously but still sequentially. We could start them to run in parallel and await `Task.WhenAll(...)`.

SzymonPobiega: Hmmm... Maybe going with multiple callbacks was a bit over-engineered and we should stick with non-additive `set`-style API? Then there would be no problems with concurrency and the list of callbacks would not be necessary.

MarcinHoppe: ❤️ 

Let's release a simple thing and get feedback.

andreasohlund:
You had me at "simple" :heart:

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateLazyNativeTransactionSynchronizedStorageSession.cs
```diff
@@ -22,23 +23,28 @@ public NHibernateLazyNativeTransactionSynchronizedStorageSession(Func<ISession>
...
+        {
+            callbacks.Add(callback);
+        }
+
         public ITransaction Transaction => Session.Transaction;
 
-        public Task CompleteAsync()
+        public async Task CompleteAsync()
         {
+            await callbacks.InvokeAll().ConfigureAwait(false);
```
tmasternak: What if callbacks throw?

SzymonPobiega:
I've simplified the API by removing the possibility for registering multiple callbacks and added unit tests to prove that session is not committed when callback throws an exception.

src/NServiceBus.NHibernate/Outbox/NHibernateOutboxTransaction.cs
```diff
@@ -25,17 +25,23 @@ public void Dispose()
...
+                throw new Exception("Save changes callback for this session has already been registered.");
+            }
+            onSaveChangesCallback = callback;
         }
 
         public async Task Commit()
         {
-            await callbacks.InvokeAll().ConfigureAwait(false);
-
+            if (onSaveChangesCallback != null)
```
MarcinHoppe: I think you can get rid of this `if` statement using the elvis operator (`?.`) and the `Invoke` method.

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateAmbientTransactionSynchronizedStorageSession.cs
```diff
@@ -36,7 +40,10 @@ public void Dispose()
 
         public async Task CompleteAsync()
         {
-            await callbacks.InvokeAll().ConfigureAwait(false);
+            if (onSaveChangesCallback != null)
```
MarcinHoppe: Ditto.

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateLazyNativeTransactionSynchronizedStorageSession.cs
```diff
@@ -23,16 +23,23 @@ public NHibernateLazyNativeTransactionSynchronizedStorageSession(Func<ISession>
...
+            }
+            onSaveChangesCallback = callback;
         }
 
         public ITransaction Transaction => Session.Transaction;
 
         public async Task CompleteAsync()
         {
-            await callbacks.InvokeAll().ConfigureAwait(false);
+            if (onSaveChangesCallback != null)
```
MarcinHoppe: Same here.

src/NServiceBus.NHibernate/Outbox/NHibernateOutboxTransaction.cs
SzymonPobiega: Wouldn't this return `null` as a result and then `await` would blow up?

MarcinHoppe: You are right, I forgot about the async part. Ignore my previous comment.

SzymonPobiega:
@tmasternak can you give it another try?

src/NServiceBus.NHibernate/Outbox/NHibernateOutboxTransaction.cs
```diff
@@ -22,12 +23,25 @@ public void Dispose()
...
+            {
+                throw new Exception("Save changes callback for this session has already been registered.");
+            }
+            onSaveChangesCallback = callback;
+        }
+
+        public async Task Commit()
+        {
+            if (onSaveChangesCallback != null)
```
tmasternak: Do we need to call `Transaction.Dispose` even if callback throws?

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateAmbientTransactionSynchronizedStorageSession.cs
```diff
@@ -20,29 +21,34 @@ public NHibernateLazyAmbientTransactionSynchronizedStorageSession(Func<IDbConnec
...
 
-        public Task CompleteAsync()
+        public async Task CompleteAsync()
         {
-            return Task.FromResult(0);
+            if (onSaveChangesCallback != null)
+            {
+                await onSaveChangesCallback(this).ConfigureAwait(false);
+            }
+            if (session.IsValueCreated)
```
tmasternak: Same question as above about `session.Value.Dispose()`

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateLazyNativeTransactionSynchronizedStorageSession.cs
```diff
@@ -9,10 +9,11 @@
     class NHibernateLazyNativeTransactionSynchronizedStorageSession : CompletableSynchronizedStorageSession, INHibernateSynchronizedStorageSession
     {
         Lazy<ISession> session;
+        Func<SynchronizedStorageSession, Task> onSaveChangesCallback;
 
         public NHibernateLazyNativeTransactionSynchronizedStorageSession(Func<ISession> sessionFactory)
         {
-            session =new Lazy<ISession>(() =>
+            session = new Lazy<ISession>(() =>
```
tmasternak: Do we need `s` declaration in labmda? Maybe just `return ...`

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateLazyNativeTransactionSynchronizedStorageSession.cs
```diff
@@ -22,23 +23,35 @@ public NHibernateLazyNativeTransactionSynchronizedStorageSession(Func<ISession>
...
+
         public ITransaction Transaction => Session.Transaction;
 
-        public Task CompleteAsync()
+        public async Task CompleteAsync()
         {
+            if (onSaveChangesCallback != null)
+            {
+                await onSaveChangesCallback(this).ConfigureAwait(false);
+            }
```
tmasternak: Same question about `Dispose`

src/NServiceBus.NHibernate/Outbox/NHibernateOutboxTransaction.cs
SzymonPobiega: Added disposing off of the transaction to `Dispose`

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateLazyNativeTransactionSynchronizedStorageSession.cs
SzymonPobiega: Added disposing off of the transaction to `Dispose`

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateAmbientTransactionSynchronizedStorageSession.cs
SzymonPobiega: I think this is fine as the session is used in a `using` block and the dispose will take care of `session.Value`.

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateLazyNativeTransactionSynchronizedStorageSession.cs
SzymonPobiega: Now that I store the result of `BeginTransaction` it can't be done ;-)

SzymonPobiega:
@tmasternak I've fixed what I could ;-)

PR #267, opened 2017-02-23
Incorrect schema creation for timeout entity causes performance degradation

MarcinHoppe:
### Who's affected

Users of NHibernate persistence who pass their own NHibernate `Configuration` object i.e.

```
var configuration = new Configuration();

configuration.SetProperty(Environment.Dialect, "NHibernate.Dialect.MsSql2012Dialect");
configuration.SetProperty(Environment.ConnectionString, @"...");

config.UsePersistence<NHibernatePersistence>().UseConfiguration(configuration);
```

### Symptoms

The problem manifests itself in wrong schema generated by the installers. Depending on the container used, there might be two variations of symptoms:

1. The schema for the *all* the tables is incorrect. All the primary keys are non-clustered (only `TimeoutEntity` table should have a non-clustered PK)
2. The schema for the `TimeoutEntity` has a clustered PK (should be non-clustered) and the `TimeoutEntity_EndpointIdx` index has wrong column order

### Root cause

When passing an external configuration object, it is shared by all the persistence concerns and all the mappings are registered in it. The first installer that runs (the order depends on the container/version used), creates the schema for *all the persistence concerns*. The timeout installer uses a specialized schema creation class. Depending on the order, this specialized class can either
1. be used for all tables
2. be not used at all

### Proposed solution

1. Make sure all the installers use the specialized schema tool (`OptimizedSchemaUpdate`)
2. Make sure the substitutions done in `OptimizedSchemaUpdate` are only executed in context of the timeouts table
3. Log a warning if incorrect schema is detected


src/NServiceBus.NHibernate.AcceptanceTests-Oracle/ConfigureNHibernatePersistence.cs
```diff
@@ -13,11 +13,11 @@ public void Configure(BusConfiguration config)
     {
         NHibernateSettingRetriever.AppSettings = () => new NameValueCollection
         {
-            {"NServiceBus/Persistence/NHibernate/connection.driver_class", "NHibernate.Driver.OracleDataClientDriver"},
+            {"NServiceBus/Persistence/NHibernate/connection.driver_class", "NHibernate.Driver.OracleManagedDataClientDriver"},
```
ramonsmits: Is there a specific reason for changing to another oracle driver? The reason is missing in the commit.

src/NServiceBus.NHibernate/TimeoutPersisters/IncorrectIndexDetector.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            try
+            {
+                connectionHelper.Prepare();
+                var connection = connectionHelper.Connection;
+
+                var index = GetIndex(connection, timeoutEntityMapping, "TimeoutEntity_EndpointIdx");
+                if (index == null)
+                {
+                    Logger.Warn("Could not find TimeoutEntity_EndpointIdx index definition. Consult NServiceBus NHibernate persistence documentation for details on how to create this index.");
```
ramonsmits: @MarcinHoppe Shouldn't we include what this can cause? Something like 

> Index 'TimeoutEntity_EndpointIdx' is missing. This causes severe performance degradation in timeout management. Consult the NServiceBus NHibernate persistence documentation on how to create this index.

ramonsmits: @MarcinHoppe Shouldn't we maybe just write this as log level ERROR to indicate that it very severe?

src/NServiceBus.NHibernate/TimeoutPersisters/IncorrectIndexDetector.cs
```diff
@@ -0,0 +1,120 @@
...
+                    return;
+                }
+
+                var indexCorrect = index.ColumnsCount == 2
+                    && index.ColumnNameAt(1) == nameof(TimeoutEntity.Endpoint)
+                    && index.ColumnNameAt(2) == nameof(TimeoutEntity.Time);
+
+                if (!indexCorrect)
+                {
+                    Logger.Warn("The TimeoutEntity_EndpointIdx index has incorrect column order. Consult NServiceBus NHibernate persistence documentation for details on how to create this index.");
```
ramonsmits: @MarcinHoppe Basically the same comment.

src/NServiceBus.NHibernate/TimeoutPersisters/OptimizedSchemaUpdate.cs
```diff
@@ -101,17 +105,27 @@ public void Execute(Action<string> scriptAction, bool doUpdate)
...
-                    if (dialectScopes.Contains(dialect.GetType().FullName))
+                    if (dialectScopes.Contains(dialect.GetType().FullName) && shouldOptimizeTimeoutEntityIndexes)
                     {
-                        sql = sql.Replace("primary key (Id)", "primary key nonclustered (Id)");
-                        sql = sql.Replace("create index TimeoutEntity_EndpointIdx on TimeoutEntity (Time, Endpoint)", "create clustered index TimeoutEntity_EndpointIdx on TimeoutEntity (Endpoint, Time)");
+                        var qualifiedTimeoutEntityTableName = timeoutEntityMapping.Table.GetQualifiedName(dialect);
+
+                        if (sql.StartsWith($"create table {qualifiedTimeoutEntityTableName}"))
+                        {
+                            sql = sql.Replace("primary key (Id)", "primary key nonclustered (Id)");
```
ramonsmits: @MarcinHoppe Refactor `sql` to `sqlline` for readability?

src/NServiceBus.NHibernate/TimeoutPersisters/OptimizedSchemaUpdate.cs
```diff
@@ -101,17 +105,27 @@ public void Execute(Action<string> scriptAction, bool doUpdate)
...
+                        }
+                        else if (sql.StartsWith("create index TimeoutEntity_EndpointIdx"))
```
ramonsmits: @MarcinHoppe Might we want to make TimeoutEntity_EndpointIdx a constant that is also used in the mapping?

src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutEntity.cs
```diff
@@ -29,6 +29,11 @@ class TimeoutEntity
         public virtual byte[] State { get; set; }
 
         /// <summary>
+        /// Timeout endpoint name.
+        /// </summary>
+        public virtual string Endpoint { get; set; }
```
ramonsmits: @MarcinHoppe Add code comment that property order is used for index column order?

src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutEntityMap.cs
```diff
@@ -28,13 +28,13 @@ public TimeoutEntityMap()
...
             });
             Property(p => p.SagaId, pm => pm.Index("TimeoutEntity_SagaIdIdx"));
-            Property(p => p.Time, pm => pm.Index("TimeoutEntity_EndpointIdx"));
-            Property(p => p.Headers, pm => pm.Type(NHibernateUtil.StringClob));
             Property(p => p.Endpoint, pm =>
             {
                 pm.Index("TimeoutEntity_EndpointIdx");
                 pm.Length(440);
             });
+            Property(p => p.Time, pm => pm.Index("TimeoutEntity_EndpointIdx"));
```
ramonsmits: @MarcinHoppe Does this also affect index column order? If so, lets add a comment here too.

src/NServiceBus.NHibernate.AcceptanceTests-Oracle/ConfigureNHibernatePersistence.cs
MarcinHoppe: The reason is that our TeamCity agent's don't the have native client installed. All other branches use the managed driver.

src/NServiceBus.NHibernate/TimeoutPersisters/IncorrectIndexDetector.cs
MarcinHoppe: @ramonsmits My understanding of log levels is the following:

- WARNING: message processed successfully but the condition needs attention,
- ERROR: message processing failed but the endpoint is still alive,
- FATAL: endpoint shutdown.

Does this sound OK? 

I will include information about performance degradation in the log message.

src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutEntityMap.cs
MarcinHoppe: @ramonsmits I don't think it does, I changed it for consistency.

src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutEntity.cs
MarcinHoppe: @ramonsmits Good point, I will add it.

src/NServiceBus.NHibernate/TimeoutPersisters/OptimizedSchemaUpdate.cs
MarcinHoppe: @ramonsmits Good point, I'll refactor.

src/NServiceBus.NHibernate/TimeoutPersisters/OptimizedSchemaUpdate.cs
MarcinHoppe: @ramonsmits Good point, I will change it.

MarcinHoppe:
@ramonsmits Can you take a look at it again (this time with Oracle detection included? If this looks good, I will rebase, squash and cherry-pick to develop and 7.x branches.

MarcinHoppe:
@SzymonPobiega Can you give this PR a glance?

SzymonPobiega:
@MarcinHoppe I'll do

src/NServiceBus.NHibernate/TimeoutPersisters/IncorrectIndexDetector.cs
SzymonPobiega: 👍 for `WARNING`

src/NServiceBus.NHibernate/TimeoutPersisters/IncorrectIndexDetector.cs
```diff
@@ -0,0 +1,164 @@
...
+            {
+                var restrictions = new string[5]
+                {
+                    entity.Table.Catalog,
+                    entity.Table.Schema,
+                    entity.Table.Name,
+                    indexName,
+                    null
+                };
+                return new SqlServerIndex(connection.GetSchema("IndexColumns", restrictions));
```
SzymonPobiega: Nitpicking. Since in this path we already know we are in "SQL Server" zone, could we here transform the `DataTable` to a `string[]` containing the names of the columns? Then we would not need `IIndex` at all.

MarcinHoppe: The point is that we also need the position of column in the index, not just the name. Where this information is stored is also different between SQL Server and Oracle.

SzymonPobiega: 👍

src/NServiceBus.NHibernate/TimeoutPersisters/IncorrectIndexDetector.cs
ramonsmits: Those definitions make sense once the endpoint has started and is actually processing messages but this is part of startup I'm fine with WARN as the level but are we never logging stuff as ERROR during startup?

MarcinHoppe: IMO ERROR should be logged at startup if the condition is severe enough to prevent endpoint from starting.

SzymonPobiega:
@MarcinHoppe ready for merging?

MarcinHoppe:
Merge yes, release no. We want to release this together with 7.0.6 and 7.1.4. In order to do that, we need to have guidance on how to manually get these indexes correct.

Also, 7.0.6 and 7.1.4 hotfixes have not been tested with customer-provided repro yet.

PR #271, opened 2017-03-19
subscription cache should not be static

SimonCropp:


SzymonPobiega:
Thanks @SimonCropp ❤️

Issue #280, opened 2017-04-21
Unique constraints are not generated for saga correlation properties


tmasternak:
## Problem

When using common base class for saga data (e.g. `ContainSagaData`) auto-generated saga tables will have unique constraint set for correlation property only if:
  * Saga data type has been scanned first
  * Saga data type correlation property has the same name as the correlation property of the saga data type mapped first

## Who's affected

All users who have at least two sagas sharing a base class (e.g. `ContainSagaData`) with different correlation property names.

## Symptoms

In concurrent message processing scenarios it is possible to end-up with saga duplicates. Two messages start the same logical saga and both succeed as there is no unique constraint on correlation property.

## Affected versions

Version 7.0.0 and later are affected.

## Root cause

NHibernate wrongly associates saga data type with `SagaMetadata` instance in [SagaModelMapper](https://github.com/Particular/NServiceBus.NHibernate/blob/develop/src/NServiceBus.NHibernate/SagaPersisters/SagaModelMapper.cs#L251) as a side effect of traversing inheritance hierarchy up and down. The end result is that all saga data types are associated with the same `SagaMetadata` instance which is later used for [setting uniqueness constraint](https://github.com/Particular/NServiceBus.NHibernate/blob/develop/src/NServiceBus.NHibernate/SagaPersisters/SagaModelMapper.cs#L166).

### Backported to

- [7.2.1](https://github.com/Particular/NServiceBus.NHibernate/releases/tag/7.2.1)
- [7.1.6](https://github.com/Particular/NServiceBus.NHibernate/releases/tag/7.1.6)

jochenz:
Thanks for your swift detection of the root cause. Looking forward to the patch.

PR #282, opened 2017-04-24
Use optimized schema update to ensure correct indexes are created

SzymonPobiega:


SzymonPobiega:
I squashed it to be ready to cherry pick to the hotfix branch.

SzymonPobiega:
Also, I narrowed down the affected specification in #280 to cases where the sagas share a base class. In case they don't share base class (e.g. implement `IContainSagaDataDirectly`) there is no problem.

SzymonPobiega:
Replaced by #283

tmasternak:
Should we delete the branch?

Issue #278, opened 2017-04-13
TimeoutEntity clustered index not created as documented


cremor:
According to the [documentation](https://docs.particular.net/nservicebus/upgrades/nhibernate-timeouts-index) and the [version 7.1.4 pull request](https://github.com/Particular/NServiceBus.NHibernate/pull/270) the TimeoutEntity table should have a nonclustered primary key.

But if I start the [Simple NHibernate Persistence Usage](https://docs.particular.net/samples/nhibernate/simple/) sample the created table does have a clustered primary key:

    CREATE TABLE [dbo].[TimeoutEntity] (
        [Id]          UNIQUEIDENTIFIER NOT NULL,
        [Destination] NVARCHAR (1024)  NULL,
        [SagaId]      UNIQUEIDENTIFIER NULL,
        [State]       VARBINARY (MAX)  NULL,
        [Endpoint]    NVARCHAR (440)   NULL,
        [Time]        DATETIME         NULL,
        [Headers]     NVARCHAR (MAX)   NULL,
        PRIMARY KEY CLUSTERED ([Id] ASC)
    );
    GO
    CREATE NONCLUSTERED INDEX [TimeoutEntity_SagaIdIdx]
        ON [dbo].[TimeoutEntity]([SagaId] ASC);
    GO
    CREATE NONCLUSTERED INDEX [TimeoutEntity_EndpointIdx]
        ON [dbo].[TimeoutEntity]([Endpoint] ASC, [Time] ASC);

In case this is dependend on the database version: I tested this using SQL Server LocalDB 13.0 (`Data Source=(localdb)\MSSQLLocalDB`)

### Backported to

- [7.2.1](https://github.com/Particular/NServiceBus.NHibernate/releases/tag/7.2.1)
- [7.1.6](https://github.com/Particular/NServiceBus.NHibernate/releases/tag/7.2.6)

MarcinHoppe:
@cremor Thanks for reporting this. I managed to reproduce the problem.

MarcinHoppe:
@cremor This issue was fixed in 7.2.1 and 7.1.6 patches.

PR #284, opened 2017-04-25
Unique constraints are not generated for saga correlation properties

SzymonPobiega:
Fixed appication of Unique constraing on saga correlation properties
Use optimized schema update to ensure correct indexes are created.


MarcinHoppe:
Fixes #280

MarcinHoppe:
Fixes #278

PR #285, opened 2017-04-26
Script generator usable in a test

SzymonPobiega:
Adds `ScriptGenerator` class that can be used in tests to generate table creation scripts.

The `DDL.cs` test contains an example of how this would be used by users to generate their schemas as part of the build process.


SzymonPobiega:
@Particular/nhibernate-persistence-maintainers this seems to me a low hanging fruit that can drastically improve the deployment/scripting story. Please take a look, it's been a while since this has been done.

mauroservienti:
NIce! does this mean that users can generate scripts during build and use them to deploy schema to environments?

MarcinHoppe:
I think this would be best split into 2 parts: expose the necessary bits from the persister and add doco / sample how to incorporate it into your own project to generate the scripts. Thoughts?

SzymonPobiega:
@mauroservienti yeah. @MarcinHoppe I agree it needs a sample but I would leave the `GenerateSciptsTest.cs` also here to make sure everything works as expected.

SzymonPobiega:
I wonder if we would have to allow access to our build server to show it end-to-end...

PR #289, opened 2017-05-05
Fixed not awaited task in CachedSubscriptionPersister

SzymonPobiega:


src/NServiceBus.NHibernate/Subscriptions/CachedSubscriptionPersister.cs
```diff
@@ -19,16 +19,14 @@ public CachedSubscriptionPersister(ISessionFactory sessionFactory, TimeSpan expi
 
         public override Task Subscribe(Subscriber subscriber, MessageType messageType, ContextBag context)
         {
-            base.Subscribe(subscriber, messageType, context);
             cache.Clear();
-            return Task.FromResult(0);
+            return base.Subscribe(subscriber, messageType, context);
```
MarcinHoppe: Is the sequence relevant here? For example, previously a failed subscribe (exception) would not clear the cache and now it does. Is this intentional?

src/NServiceBus.NHibernate/Subscriptions/CachedSubscriptionPersister.cs
```diff
@@ -19,16 +19,14 @@ public CachedSubscriptionPersister(ISessionFactory sessionFactory, TimeSpan expi
...
         }
 
         public override Task Unsubscribe(Subscriber address, MessageType messageType, ContextBag context)
         {
-            base.Unsubscribe(address, messageType, context);
             cache.Clear();
-            return Task.FromResult(0);
+            return base.Unsubscribe(address, messageType, context);
```
MarcinHoppe: Same question here.

src/NServiceBus.NHibernate/Subscriptions/CachedSubscriptionPersister.cs
SzymonPobiega: Not intentional but I figured it won't make things worse. Clearing the cache won't hurt much and subscribe failures should not happen that often in the stable system. I figures the code will be cleaner this way (vs using `await`).

PR #290, opened 2017-05-18
[wip] test unicode chars

weralabaj:
https://github.com/Particular/PlatformDevelopment/issues/1256

### DIFF
--- a/src/NServiceBus.NHibernate.AcceptanceTests/App_Packages/NSB.AcceptanceTests.6.1.1/Reliability/Outbox/When_headers_keys_contain_special_characters.cs
+++ b/src/NServiceBus.NHibernate.AcceptanceTests/App_Packages/NSB.AcceptanceTests.6.1.1/Reliability/Outbox/When_headers_keys_contain_special_characters.cs
@@ -55,7 +55,7 @@
                 EndpointSetup<DefaultServer>(b =>
                 {
                     b.GetSettings().Set("DisableOutboxTransportCheck", true);
-                    b.Pipeline.Register("BlowUpBeforeDispatchBehavior", new BlowUpBeforeDispatchBehavior(), "For testing");
+                    b.Pipeline.Register("BlowUpBeforeDispatchBehavior", new BlowUpBeforeDispatchBehavior(), "Force reading the message from Outbox storage.");
                     b.Recoverability().Immediate(a => a.NumberOfRetries(1));
                     b.EnableOutbox();
                 });
--- a/src/NServiceBus.NHibernate.AcceptanceTests/App_Packages/NSB.AcceptanceTests.6.1.1/Reliability/Outbox/When_headers_keys_contain_special_characters.cs
+++ b/src/NServiceBus.NHibernate.AcceptanceTests/App_Packages/NSB.AcceptanceTests.6.1.1/Reliability/Outbox/When_headers_keys_contain_special_characters.cs
@@ -11,41 +11,38 @@
 
     public class When_headers_contain_special_characters : NServiceBusAcceptanceTest
     {
-        static Guid messageId = Guid.NewGuid();
-        static bool SavedOutBoxRecord = false;
         static Dictionary<string, string> sentHeaders = new Dictionary<string, string>
-                          {
-                            { "a-B1", "a-B" },
-                            { "a-B2", "a-ɤϡ֎ᾣ♥-b" },
-                            { "a-ɤϡ֎ᾣ♥-B3", "a-B" },
-                            { "a-B4", "a-\U0001F60D-b" },
-                            { "a-\U0001F605-B5", "a-B" },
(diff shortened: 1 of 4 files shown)


src/NServiceBus.NHibernate.AcceptanceTests/App_Packages/NSB.AcceptanceTests.6.1.1/Reliability/Outbox/When_headers_keys_contain_special_characters.cs
```diff
@@ -0,0 +1,95 @@
...
+                public Task Handle(SendOrderAcknowledgement message, IMessageHandlerContext context)
+                {
+                    Context.MessageReceived = true;
+                    Context.UnicodeHeaders = (Dictionary<string, string>) context.MessageHeaders;
+                    return Task.FromResult(0);
+                }
+            }
+        }
+
+        public class Pläceörder : ICommand
```
timbussmann: not sure we want to also test special characters in class names when this test is focused on header values?

src/NServiceBus.NHibernate.AcceptanceTests/App_Packages/NSB.AcceptanceTests.6.1.1/Reliability/Outbox/When_headers_keys_contain_special_characters.cs
```diff
@@ -0,0 +1,95 @@
...
+        public async Task Outbox_should_work()
+        {
+            var context =
+                await Scenario.Define<Context>()
+                .WithEndpoint<OutboxEndpoint>(b => b.When(session => session.SendLocal(new Pläceörder())))
+                .Done(c => c.MessageReceived)
+                .Run(TimeSpan.FromSeconds(20));
+
+            Assert.IsNotEmpty(context.UnicodeHeaders);
+            CollectionAssert.IsSubsetOf(sentHeaders, context.UnicodeHeaders);
```
timbussmann: I switched over to use IsSupersetOf because it gives a more apropriate message on failure (need to be careful with the parameters though)

src/NServiceBus.NHibernate.AcceptanceTests/App_Packages/NSB.AcceptanceTests.6.1.1/Reliability/Outbox/When_headers_keys_contain_special_characters.cs
```diff
@@ -0,0 +1,95 @@
...
+            }
+
+            class SendOrderAcknowledgementHandler : IHandleMessages<SendOrderAcknowledgement>
+            {
+                public Context Context { get; set; }
+
+                public Task Handle(SendOrderAcknowledgement message, IMessageHandlerContext context)
+                {
+                    Context.MessageReceived = true;
+                    Context.UnicodeHeaders = (Dictionary<string, string>) context.MessageHeaders;
```
timbussmann: I think this test doesn't test the full story:
In this setup, the outbox will store the outgoing messages, but the message received here, won't come from the outbox as it will dispatch the message directly to the transport. The outbox would only come into play when you receive a "duplicate message". In that case it should re-dispatch the message stored in the outbox which might contain corrupted headers.

I think this could be done by sending a new PlaceOrder command after MessageReceived is set to true with the same message id as the first message.

weralabaj: >  as it will dispatch the message directly to the transport.

I was also wondering about this, I;ll double check how that works

src/NServiceBus.NHibernate.AcceptanceTests/App_Packages/NSB.AcceptanceTests.6.1.1/Reliability/Outbox/When_headers_keys_contain_special_characters.cs
weralabaj: yeah, i'll leave it out

weralabaj:
This PR is just to verify if the test will pass immediately, but it'll be incorporated as part of AcceptanceTest suite so no need to merge. Still @SzymonPobiega if you could review if this test makes sense, that will be appreciated and then we can close this PR.

SzymonPobiega:
Looks good @weralabaj. If I understand correctly, it sends a message to itself and in the handler of that message it sends another message (this time with unicode headers). Because of the special behavior the outbox dispatch will fail the first time and retry will kick in to ensure the message headers are actually loaded from the database.

weralabaj:
that's the idea :)

weralabaj:
need to verify oracle as well :D

PR #291, opened 2017-05-22
Add acceptance tests to test special character headers

timbussmann:
raising this as a PR to let the tests run on the different SQL servers. Feel free to pull this in or to close this once the tests ran (will be part of the acceptance tests package with the next version anyway)


SzymonPobiega:
well, as far as I see the `TransportOperations` column in the outbox is created as `nvarchar` so it should be fine. I'll close this and wait for it to come back in the AT package.

PR #298, opened 2017-07-18
Adds scripts for creating index on FKs when saga data contains child elements

### LINKED ISSUE #297: Deadlocks while persisting saga data collections
**Problem**
When using saga data that contains collections I'm getting quite a few deadlocks (SQL Server). The primary saga data (inheriting IContainSagaData) is persisted properly, but collections (which are stored in a seperate table) are not. It seems that these collections cause table locks thus causing deadlocks while processing a lot of saga messages (even if saga's aren't handling concurrent messages).

non-optimized

**Possible solution**
It would be nice to add a non-unique index on the foreign key, avoiding the need for a full table scan to find the proper collection data for a saga. Also I would like to add a new column, a generated guid, as a primary key. This avoids a table rescan when deleting the indirect saga data.

optimized

Relevant code: https://github.com/Particular/NServiceBus.NHibernate/blob/1c282a2aaf1804f40e227bb12647734a58443493/src/NServiceBus.NHibernate/SagaPersisters/SagaModelMapper.cs#L188

SzymonPobiega:
Resolves #297

### DIFF
--- a/src/NServiceBus.NHibernate.Tests/DDL.MySaga.approved.txt
+++ b/src/NServiceBus.NHibernate.Tests/DDL.MySaga.approved.txt
@@ -1,6 +1,19 @@
 ﻿
     
+    if exists (select 1 from sys.objects where object_id = OBJECT_ID(N'[FK236CFC11C9055018]') AND parent_object_id = OBJECT_ID('MySaga_CollectionEntry'))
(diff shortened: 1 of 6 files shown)


SzymonPobiega:
@Particular/nhibernate-persistence-maintainers please review.

Issue #304, opened 2017-09-05
Use locking hint when fetching timeouts


dvdstelt:
### Who's affected

Users who run `TransactionScope` transaction mode endpoints with transports that don't support delayed messages internally (MSMQ, SQL Server transport < 3.1)

### Symptoms

If an endpoint is scaled-out, when a delayed message matures it can be picked up by multiple instances that all compete to dispatch the message to the final destination. In `TransactionScope` mode the actual message is guaranteed to be sent only once but the "loosing" instances will log exception that the message has already been dispatched.

### Description

In NServiceBus 5.x and NServiceBus.NHibernate 6.x timeframe, we loaded timeouts using a locking hint.
The locking hint was removed while upgrading to NServiceBus Core 6 Beta packages in the PR [`Updating to core v6`](https://github.com/Particular/NServiceBus.NHibernate/commit/7006d58ab7d9e2b8f83041a867a4b048999926da)

You can see the change in [this diff](https://github.com/Particular/NServiceBus.NHibernate/commit/7006d58ab7d9e2b8f83041a867a4b048999926da#diff-98e9323d424a44d6bf97f400409b04fdL231) where we switched from `Get<TimeoutEntity>` to `session.QueryOver<TimeoutEntity>`.

The question that needs to be investigated is, should we move back to providing a locking hint or is this being dealt with properly using transactions?
Possibly we should change back, since a customer is receiving exception messages about concurrency issues related to timeout messages.

PR #305, opened 2017-09-08
[WIP] Use upgrade lock

SzymonPobiega:
Connects to #304 

Uses `SELECT FOR UPDATE` type of query when `Peek`ing timeouts to prevent multiple instances of the endpoint from attempting to dispatch a single timeout. Should lower the chances of stealing timeouts. Will only work in `TransactionScope` mode because only then there will be a transaction between `Peek` and `TryDelete`. In other transaction modes will behave similar to like it behaved before because the lock will be released prior to deleting.

### DIFF
--- a/src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutPersister.cs
+++ b/src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutPersister.cs
@@ -154,10 +154,7 @@ namespace NServiceBus.TimeoutPersisters.NHibernate
             using (var session = await OpenSession(context).ConfigureAwait(false))
             {
                 var id = Guid.Parse(timeoutId);
-                var te = session.Session().QueryOver<TimeoutEntity>()
-                    .Where(x => x.Id == id)
-                    .List()
-                    .SingleOrDefault();
+                var te = session.Session().Get<TimeoutEntity>(id, LockMode.Upgrade);
 
                 var timeout = MapToTimeoutData(te);
                 return timeout;


SzymonPobiega:
@weralabaj @dvdstelt please review.

weralabaj:
would it be good to notify @Particular/nhibernate-persistence-maintainers ? ;P

dvdstelt:
Is this still being tested, or can we merge and release? Any status update?

SzymonPobiega:
@dvdstelt I'll try to fix the build and we should be able to release. Question is, which version should we release this in?

dvdstelt:
@SzymonPobiega Of NHibernate? According [to this](https://docs.particular.net/nservicebus/upgrades/supported-versions#persistence-packages-nservicebus-nhibernate) we should update 6.2 and 7 I guess?

SzymonPobiega:
I tried to create a test for it but failed as SQLite that we use in tests just ignores the `LockMode.Upgrade`. So I used SQL Profiler to verify that the underlying SQL is in fact generated correctly and the only difference is the lock hint. 

Now the question to @Particular/nhibernate-persistence-maintainers : since we are on the verge of releasing 8, should we release 7.2 patch with this fix or just move it to 8?

dvdstelt:
Will v8 have a dependency on NSB6 or will it be NSB7? Because I do think current NServiceBus users should be able to benefit from this.

SzymonPobiega:
NSB7 only. The reason I am a bit hesitant and would like other @Particular/nhibernate-persistence-maintainers to chime in is the fact that this fix does not affect the correctness of the solution. I mean, the this we gain is less exceptions in the logs but the behavior stays the same. And the only affected configuration is MSMQ + NHibernate.

Does it justify releasing a patch, causing all other unaffected people to have to upgrade?

dvdstelt:
>  the fact that this fix does not affect the correctness of the solution

The end-result will be no different, but it puts more burden on both SQL Server and the endpoints.

> Does it justify releasing a patch, causing all other unaffected people to have to upgrade?

We 'broke' the behavior from NSB5 to NSB6, so I would say this justifies fixing this. Others will benefit from this as well.

And I don't think "causing all other unaffected people to have to upgrade" is an issue. People only upgrade if they have to (because of some feature or bugfix). Or if they are quite certain that any upgrade will work, they'll just do it when they are changing their own code anyway. Not just because of us bumping a version in our packages.

So because it's less exceptions, less logging but better behaving package inside SQL ánd it's because we were the ones who broke this behavior, I strongly vote for doing it current version.

DavidBoike:
:+1: to what Dennis said. Plus it's a trivially minor fix to implement.

SzymonPobiega:
OK, proceeding with the release today.

SzymonPobiega:
Closed via d2407060d2447b761c9457935478db08de7993cd

PR #314, opened 2017-10-30
Project cleanup

bording:
This moves the project to using Particular.Packaging and cleans up a few other things related to moving to the new project system.


src/NServiceBus.NHibernate/NServiceBus.NHibernate.csproj
```diff
@@ -1,32 +1,31 @@
...
+
+  <ItemGroup>
     <PackageReference Include="Fody" Version="2.*" PrivateAssets="All" />
     <PackageReference Include="Janitor.Fody" Version="1.*" PrivateAssets="All" />
     <PackageReference Include="Obsolete.Fody" Version="4.*" PrivateAssets="All" />
     <PackageReference Include="Particular.CodeRules" Version="0.2.0" PrivateAssets="All" />
+    <PackageReference Include="Particular.Packaging" Version="0.1.0" PrivateAssets="All" />
     <PackageReference Include="NServiceBus" Version="[7.0.0-*, 8.0.0)" />
     <PackageReference Include="NHibernate" version="[4.0.4.4000, 5.0.0)" />
     <PackageReference Include="Iesi.Collections" version="[4.0.2, 5.0.0)" />
```
bording: It's not immediately clear to me why this dependency is here, since it's an NHibernate dependency and would be picked up transitively through the NHibernate package reference. Is there something about the 4.0.2 package specifically that we need?

MarcinHoppe: I think this is a left-over from before support for transitive dependencies. Thanks for pointing this out!

bording: Would you like me to remove it as part of this PR?

MarcinHoppe: That would be perfect but I can also remove it in a separate PR later.

bording:
So it turns out that because the NHibernate package has their Iesi.Collections dependency wrong compared to the actual released versions of Iesi.Collections, we get a warning (which is treated as an error) when we remove our separate dependency on it:

```
Warning	NU1603	NHibernate 4.0.0.4000 depends on Iesi.Collections (>= 4.0.0 && < 5.0.0) but Iesi.Collections 4.0.0 was not found. An approximate best match of Iesi.Collections 4.0.0.4000 was resolved.
```
They should fix their minimum version to be 4.0.0.4000, or even better bump it to 4.0.2.

In the mean time, I'll add the dependency back to prevent the warning, and also prevent consumers of our package hitting the same problem.

bording:
NHibenate issue: https://github.com/nhibernate/nhibernate-core/issues/1430

Won't fix older packages, but if they release a new package, we can drop the dependency at that point.

bording:
@Particular/nhibernate-persistence-maintainers This PR still needs to be merged. I see that the Perf test timed out, but I don't think that is related to these changes.

SzymonPobiega:
Sorry @bording , we dropped the ball :(

PR #330, opened 2018-05-01
Hotfix 7.2.3

ramonsmits:


ramonsmits:
Based on the build history https://builds.particular.net/viewType.html?buildTypeId=NServiceBusNHibernate_315PerformanceTestsFrameworkOnWindows&tab=buildTypeHistoryList&branch_NServiceBusNHibernate=__all_branches__

- the performance tests can be ignored
- some AT's are failing but don't have anything to do with the changed code
- locally the failing tests run for both SQL and Oracle, but sometimes they fail.

PR #312, opened 2017-10-19
update to NH5

SimonCropp:

PR #341, opened 2018-06-15
Migrate to NHibernate 5.*

Scooletz:
This PR replaces #312 . 

Notes:
- commits from  #312 have been squashed to make it easier to rebase it on top of the current `develop` and put as the first commit in this PR
- several methods were turned from non-async into `*Async` counterparts (for example `.List()`, `.Get` and a few others)
- this PR moves transport into **.NET Core/Standard**
- this PR changes one accepted API, as NH 5.1 changes the call to `OBJECT_ID` for a saga. According to the doco for this function, `if object_name is varchar, it is implicitly converted to nvarchar`. I'm applying this change then
- this PR introduced a custom `DateTimeType` to skip moving the `DateTime` properties to `DATETIME2` SqlType. This requires a NH expert to take a loot at ;-)


Scooletz:
@Particular/nhibernate-persistence-maintainers Could I get a reviewer assigned for this? I unWIPped it as well.

ramonsmits:
I'm not a maintainer, but the code stuff looks great. You might want to ping @bording to review the packaging.

src/NServiceBus.NHibernate.AcceptanceTests-SqlTransport/NServiceBus.NHibernate.AcceptanceTests-SqlTransport.csproj
```diff
@@ -13,8 +17,12 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
     <PackageReference Include="NServiceBus.AcceptanceTests.Sources" Version="7.0.0" />
     <PackageReference Include="NServiceBus.SqlServer" Version="4.0.0" />
-    <PackageReference Include="NUnit" Version="3.*" />
-    <PackageReference Include="NUnit3TestAdapter" Version="3.8.0-alpha1" />
+    <PackageReference Include="NServiceBus" Version="7.0.0-*" />
+    <PackageReference Include="NServiceBus.SqlServer" Version="4.0.0-beta0004" />
+    <PackageReference Include="NServiceBus.AcceptanceTesting" Version="7.0.0-*" />
+    <PackageReference Include="NServiceBus.AcceptanceTests.Sources" Version="7.0.0-*" />
```
DavidBoike: Shouldn't these be NServiceBus (and friends) 7.* and SqlT 4.* rather than prerelease packages?

src/NServiceBus.NHibernate.Tests/API/APIApprovals.cs
```diff
@@ -1,18 +1,43 @@
...
-        public void Approve()
+        public void Approve__NET461()
         {
-            var publicApi = ApiGenerator.GeneratePublicApi(typeof(NHibernatePersistence).Assembly);
+            var publicApi = Filter(ApiGenerator.GeneratePublicApi(typeof(NHibernatePersistence).Assembly));
             TestApprover.Verify(publicApi);
         }
+#endif
+
+#if NETCOREAPP2_0
```
DavidBoike: Are the approvals different on each platform?

If it's just the TargetFrameworkAttribute getting you down, you can follow this recipe, and then the file is the same for both platforms.
https://github.com/Particular/NServiceBus.Persistence.Sql/blob/develop/src/SqlPersistence.Tests/APIApprovals.cs

src/NServiceBus.NHibernate.Tests/ApprovalFiles/DDL.GatewayDeduplication.approved.txt
```diff
@@ -4,6 +4,6 @@
     
     create table GatewayDeduplication (
         Id NVARCHAR(255) not null,
-       TimeReceived DATETIME null,
+       TimeReceived DATETIME2 null,
```
DavidBoike: None of these DATETIME -> DATETIME2 changes could be considered a breaking change, could it?

src/NServiceBus.NHibernate/NServiceBus.NHibernate.csproj
```diff
@@ -1,30 +1,28 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net452</TargetFramework>
+    <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
```
DavidBoike: I wonder if we need to specifically document that although NSB supports net452, the NHibernate dependency forces at least 461?

src/NServiceBus.NHibernate/Outbox/OutboxCleaner.cs
```diff
@@ -52,7 +52,7 @@ protected override Task OnStart(IMessageSession busSession)
             circuitBreaker = new RepeatedFailuresOverTimeCircuitBreaker(
                 "OutboxCleanupTaskConnectivity",
                 timeToWaitBeforeTriggeringCriticalError,
-                ex => criticalError.Raise("Failed to clean the Oubox.", ex)
+                ex => criticalError.Raise("Failed to clean the Outbox.", ex)
```
DavidBoike: Nice catch

Scooletz: `Ouchbox` could have been worse 😉

src/NServiceBus.NHibernate.Tests/ApprovalFiles/DDL.GatewayDeduplication.approved.txt
Scooletz: @DavidBoike I have no idea. That's a change in v5.0 of NH and it can be configured to use the previous one, but this is global configuration setting so I wouldn't do it. Do we need a test for this then?

For sake of reference, a SO question mentioning it https://stackoverflow.com/questions/32679529/nhibernate-and-sql-server-datetime2-and-supporting-both-sql-server-and-oracle

src/NServiceBus.NHibernate/NServiceBus.NHibernate.csproj
Scooletz: Isn't nuget range specification sufficient enough?

src/NServiceBus.NHibernate.Tests/ApprovalFiles/DDL.GatewayDeduplication.approved.txt
DavidBoike: Seems like datetime2 is larger range and higher precision than datetime and should accept datetime values just fine. Think we're OK.

DavidBoike:
The SQL + SQL Server Tests failed on a silly timeout test:

> Expected: greater than or equal to 00:00:02
>  But was:  00:00:01.9869452

I can't express how much I hate that test. I restarted that test, hopefully it will succeed this time.

src/NServiceBus.NHibernate.AcceptanceTests/NServiceBusAcceptanceTest.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace NServiceBus.AcceptanceTests
+{
+    public partial class NServiceBusAcceptanceTest
+    {
+        static NServiceBusAcceptanceTest()
```
bording: Why did you need to add this? The class as included in the source package already has this workaround. Was there some reason it needed to be done sooner?

src/NServiceBus.NHibernate.Tests/ApprovalFiles/DDL.GatewayDeduplication.approved.txt
bording: Is there any problem here around the fact that people will already have tables defined with `DATETIME` and now we'll be expecting `DATETIME2` instead?

src/NServiceBus.NHibernate/NServiceBus.NHibernate.csproj
```diff
@@ -1,30 +1,25 @@
...
     <PackageReference Include="Particular.CodeRules" Version="0.2.0" PrivateAssets="All" />
     <PackageReference Include="Particular.Packaging" Version="0.1.0" PrivateAssets="All" />
-    <PackageReference Include="NServiceBus" Version="[7.0.0, 8.0.0)" />
-    <PackageReference Include="NHibernate" version="[4.0.4.4000, 5.0.0)" />
-    <PackageReference Include="Iesi.Collections" version="[4.0.2, 5.0.0)" />
+    <PackageReference Include="Fody" Version="3.0.4" PrivateAssets="All" />
+    <PackageReference Include="Janitor.Fody" Version="1.6.1" PrivateAssets="All" />
+    <PackageReference Include="Obsolete.Fody" Version="4.4.0" PrivateAssets="All" />
+    <PackageReference Include="NServiceBus" Version="[7.0.1, 8.0.0)" />
+    <PackageReference Include="NHibernate" version="[5.1.0, 6.0.0)" />
```
bording: Any reason to not set the minimum to the latest package version that is out now, 5.1.3? That way users will be getting bug fixes.

bording:
I also happened to notice that the `When_adding_outbox_messages` test is still using `transaction.Commit` and `SingleOrDefault` instead of `SingleOrDefaultAsync`. Looks like there might be more opportunities to convert to async still.

bording:
Before this is merged, the TC setup needs to be modified so all tests are being run under .NET Core on Windows and Linux.

src/NServiceBus.NHibernate/NServiceBus.NHibernate.csproj
Scooletz: @bording They should get it according to their "bump up" policy. My reasoning was, that we want to always specify the lowest possible range. Setting `5.1.3` as the lower boundary would make me, as a user, think what's specific about 5.1.3

src/NServiceBus.NHibernate.AcceptanceTests/NServiceBusAcceptanceTest.cs
Scooletz: Huh, this is from the original PR that I based my work on. I could try to remove it @bording

Scooletz:
I'll review the tests in search of asyncification possibilities. I'll take a look at TC setup

src/NServiceBus.NHibernate/Outbox/OutboxCleaner.cs
```diff
@@ -60,53 +60,43 @@ protected override Task OnStart(IMessageSession busSession)
...
-            {
-                circuitBreaker.Failure(ex);
-            }
-            finally
+            while (ct.IsCancellationRequested == false)
             {
                 try
                 {
-                    cleanupTimer.Change(frequencyToRunDeduplicationDataCleanup, Timeout.InfiniteTimeSpan);
+                    await outboxPersister.RemoveEntriesOlderThan(DateTime.UtcNow - timeToKeepDeduplicationData).ConfigureAwait(false);
```
Scooletz: What was the logic of the cleaner before? Did it stop after the first usage? This changes it to an asynchronous poller with a fixed delay for every spin `frequencyToRunDeduplicationDataCleanup`

src/NServiceBus.NHibernate.AcceptanceTests/NServiceBusAcceptanceTest.cs
bording: If it works without it, let's get rid of it. That also cleans up the extra bits in the other test project files.

src/NServiceBus.NHibernate/NServiceBus.NHibernate.csproj
bording: bump up policy?

Scooletz: Should we be responsible for making users' NHibernate upgraded to the latest or should it be users' responsiblity? I think it's the latter.

bording: I don't think we need to update the package all the time, but I see no reason to release with an older version than the latest that is out when we release, especially given the nature of transitive references. They might not even have a top-level package reference, so they'd be stuck on 5.1.0.

src/NServiceBus.NHibernate/ScriptGenerator.cs
```diff
@@ -112,6 +113,7 @@ static string GenerateScript(Configuration config)
             var formatter = FormatStyle.Ddl.Formatter;
             var script = new StringBuilder();
             export.Create(s => script.Append(formatter.Format(fixUpHelper.FixUp(s))), false);
+            script.Replace(global::System.Environment.NewLine, "\r\n");
```
bording: Why has this been done? We shouldn't be forcing people to use Windows line endings if they aren't on Windows.

Scooletz: This has been done to make tests independent on the environment and it has been applied in wrong place. Will move it to the test suite

Scooletz:
@bording @Particular/nhibernate-persistence-maintainers this is ready for the review.

The checks are not green as there have been a lot of changes in TeamCity build configurations. This includes changing names, moving tests between configs and archiving some of them.

src/NServiceBus.NHibernate/NServiceBus.NHibernate.csproj
```diff
@@ -1,30 +1,25 @@
...
-    <PackageReference Include="NHibernate" version="[4.0.4.4000, 5.0.0)" />
-    <PackageReference Include="Iesi.Collections" version="[4.0.2, 5.0.0)" />
+    <PackageReference Include="NServiceBus" Version="[7.0.1, 8.0.0)" />
+    <PackageReference Include="NHibernate" version="[5.1.3, 6.0.0)" />
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="SourceLink.Create.GitHub" Version="2.5.0" PrivateAssets="All" />
+    <PackageReference Include="SourceLink.Create.GitHub" Version="2.6.0" PrivateAssets="All" />
     <DotNetCliToolReference Include="dotnet-sourcelink-git" Version="2.5.0" />
```
bording: The DotNetCliToolReference version should match the PackageReference.

src/NServiceBus.NHibernate.AcceptanceTests-SqlTransportTests/NServiceBus.NHibernate.AcceptanceTests-SqlTransportTests.csproj
```diff
@@ -1,8 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net452</TargetFramework>
+    <TargetFrameworks>net461;netcoreapp2.0</TargetFrameworks>
     <RootNamespace>NServiceBus.NHibernate.AcceptanceTests.SqlTransport</RootNamespace>
```
bording: This shouldn't be required. It would be fine to let the name derived from the project be the default.

src/NServiceBus.NHibernate.AcceptanceTests-SqlTransportTests/NServiceBus.NHibernate.AcceptanceTests-SqlTransportTests.csproj
```diff
@@ -11,10 +13,10 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
-    <PackageReference Include="NServiceBus.AcceptanceTests.Sources" Version="7.0.0" />
     <PackageReference Include="NServiceBus.SqlServer" Version="4.0.0" />
```
bording: Should this be updated to 4.1.0?

bording:
@Scooletz I noticed that you merged develop into this branch. In general, I think that's something we should avoid doing if possible, and instead feature/PR branches should be rebased.

Scooletz:
> I noticed that you merged develop into this branch. In general, I think that's something we should avoid doing if possible, and instead feature/PR branches should be rebased.

@bording I didn't want to prune earlier comments, which would be lost if I rebased it. Additionally, in a PR like this, I'd prefer to squash it before merging. Then, it makes no difference.

src/NServiceBus.NHibernate.Tests/GenerateScriptsTest.cs
```diff
@@ -45,13 +45,16 @@ public void GatewayDeduplication()
             TestApprover.Verify(script);
         }
 
+#if !NETCOREAPP2_0
+        // This test is ignored for .NETCore because of the unstable foreing key name generation in NHibernate https://github.com/nhibernate/nhibernate-core/issues/1769
```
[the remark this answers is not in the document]
Scooletz: I'll go with `!NETCOREAPP`.

src/NServiceBus.NHibernate.AcceptanceTests-SqlTransportTests/NServiceBus.NHibernate.AcceptanceTests-SqlTransportTests.csproj
Scooletz: I left the namespace unchanged (project name has been changet to get picked up by build configs)

bording:
> @bording I didn't want to prune earlier comments, which would be lost if I rebased it. Additionally, in a PR like this, I'd prefer to squash it before merging. Then, it makes no difference.

I wasn't sure if this was going to be squashed or not, which is why I tend to prefer rebasing. As long as it is indeed going to be squashed, then yes the merge doesn't matter!

src/NServiceBus.NHibernate.AcceptanceTests-SqlTransportTests/NServiceBus.NHibernate.AcceptanceTests-SqlTransportTests.csproj
bording: The reason I was suggesting to remove the line is that if you look at all the files in the project, none of them are in the `NServiceBus.NHibernate.AcceptanceTests.SqlTransport` namespace, so there's no real reason to care about maintaining that value.

The new default value based on the project name is just as good, so it seems better to remove an unneeded line from the project file!

PR #350, opened 2018-11-13
Add indexes on all collection items not only with Id properties

ramonsmits:
## Overview
This PR changes the saga data script generation. It makes sure that for child collections that do not specify an `id` property a unique index gets created on foreign key column pointing to the master table.

The resulting script was tested on Ms SQL and Oracle 19c.

<details>
<summary>Old content</summary>
Based on a support case and [issue comment](https://github.com/Particular/NServiceBus.NHibernate/issues/297#issuecomment-434366580) where a customer expected to see indexes on the FK for a child collection. Not having indexes created is likely to result in deadlocks because of table locks on the child collection table on Oracle and MSSQL.

Not heavily tested! Might need some additional test scenarios! 
</details>


tmasternak:
@DavidBoike fyi. For some reason, github is not showing the newest build status. All are green.

DavidBoike:
I think it's because this PR is so old that TC was evaluating it as `350` instead of `pull/350/merge` and thus not reporting back at all. I moved the branch protection aside (I'm a GH admin) merged, and then restored the protections.

PR #356, opened 2019-01-09
support72 update dev tooling

seanfarmar:
This updates development tooling on the `support7.2` branch.

Note: It is okay for this to be squashed.


bording:
Dealing with some TC issues, so this PR has been replaced: #359

PR #362, opened 2019-07-09
Fixing TimeoutManager index validator for quoted schemas

### LINKED ISSUE #349: TimeoutEntity_EndpointIdx validation error due to braces "[" and "]" used in schema name, even though correct index defined.
Hi, had a chat with Andreas today about "TimeoutEntity_EndpointIdx" index column validation error on the TimeoutEntity table. Because my schema contains braces when provided to the NHibernate config because otherwise it would not be recognized as a good identifier due to the use of a hyphen, the query to determine the columns used in the index does not return any results.

nHibernateConfig.SetProperty(NHibernate.Cfg.Environment.DefaultSchema, "[IOCDS-Async_NServiceBus_Scheduling]");

**As a result the following log entries are created:**

2018-11-07 14:58:38,212 [10] DEBUG NServiceBus.TimeoutPersisters.NHibernate.Installer.IncorrectIndexDetector - Detected TimeoutEntity_EndpointIdx (, )

2018-11-07 14:58:38,213 [10] WARN  NServiceBus.TimeoutPersisters.NHibernate.Installer.IncorrectIndexDetector - The TimeoutEntity_EndpointIdx index has incorrect column order. This may cause significant performance degradation of message deferral. Consult NServiceBus NHibernate persistence documentation for details on how to create this index.

**Possibly this issue can be fixed here within the GetIndex method:**
https://github.com/Particular/NServiceBus.NHibernate/blob/21b64e933de12f05486384ab27558bca083564f6/src/NServiceBus.NHibernate/TimeoutPersisters/IncorrectIndexDetector.cs

**By changing the GetIndex code as follows (only showing a snippet of the entire body):**

```
if (typeof(MsSql2005Dialect).IsAssignableFrom(dialect.GetType())) 
{ 
    var restrictions = new string[5] 
    { 
        entity.Table.Catalog, 
        entity.Table.Schema
                            .TrimStart('[')
                            .TrimEnd(']'), 
        entity.Table.Name, 
        indexName, 
        null 
    }; 
    return new SqlServerIndex(connection.GetSchema("IndexColumns", restrictions)); 
} 
```

This fix targets the use of braces with the schema specifically, perhaps the other properties should also be checked in a similar way?

Regards,
Roel

tmasternak:
Fixes #349

### DIFF
--- a/src/NServiceBus.NHibernate.Tests/TimeoutPersister/When_checking_db_indexes.cs
+++ b/src/NServiceBus.NHibernate.Tests/TimeoutPersister/When_checking_db_indexes.cs
@@ -1,0 +1,123 @@
+﻿namespace NServiceBus.NHibernate.Tests.TimeoutPersister
+{
+    using System.Data.SqlClient;
+    using System.Threading.Tasks;
+    using global::NHibernate.Cfg;
+    using global::NHibernate.Dialect;
+    using global::NHibernate.Mapping.ByCode;
+    using global::NHibernate.Tool.hbm2ddl;
+    using NUnit.Framework;
+    using TimeoutPersisters.NHibernate.Config;
+    using TimeoutPersisters.NHibernate.Installer;
+
+    class When_checking_db_indexes
+    {
+        SchemaExport schemaExport;
+
+        string dbSchemaName;
+        bool dbSchemaNeedsQuoting;
+
+        [Test]
+        public async Task Should_detect_existing_TimeoutEntity_index_in_default_configuration()
+        {
+            var configuration = await CreateTimeoutManagerObjects();
+
+            var validationResult = new TimeoutsIndexValidator(configuration).Validate();
+
+            Assert.IsTrue(validationResult.IsValid, "Validation should pass for default db structure.");
+        }
+
+        [Test]
+        public async Task Should_detect_missing_TimeoutEntity_index_in_unquoted_schema()
+        {
+            dbSchemaName = "some_schema";
+
+            var configuration = await CreateTimeoutManagerObjects();
+            await DropIndex();
+
+            var validationResult = new TimeoutsIndexValidator(configuration).Validate();
+
+            Assert.IsFalse(validationResult.IsValid, "Validation should fail if the index is missing.");
+        }
+
+        [Test]
+        public async Task Should_detect_existing_TimeoutEntity_index_in_quoted_schema()
+        {
+            dbSchemaName = "quoted-schema";
+            dbSchemaNeedsQuoting = true;
+
+            var configuration = await CreateTimeoutManagerObjects();
+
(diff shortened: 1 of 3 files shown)


src/NServiceBus.NHibernate/TimeoutPersisters/TimeoutsIndexValidator.cs
```diff
@@ -75,19 +89,33 @@ public void LogWarningIfTimeoutEntityIndexIsIncorrect()
...
             if (typeof(MsSql2005Dialect).IsAssignableFrom(dialect.GetType()))
             {
-                var restrictions = new string[5]
+                string EnsureUnquoted(string name)
                 {
-                    entity.Table.Catalog,
-                    entity.Table.Schema,
-                    entity.Table.Name,
-                    indexName,
+                    if (null == name || name.Length < 2)
```
DavidBoike: I'm concerned there could be a hidden bug in all this math and substringing. Did you consider `Regex.Replace(name, @"[\[\]]+", "")`? (Just removes any `[` or `]` anywhere within the string.) Or does it have to be only on the boundaries to be removed?

tmasternak: So what this does is the following:
  * It **does not** validate if the identifier is correct
  * If the identifier is wrapped in `[]` it assumes that it's quoted and:
   * Removes the wrapping `[]` brackets
   * Replaces all double right brackets `]]` with a single one `]`. This is needed because right brackets need to be escaped in quoted identifiers.

This is a less generic version of (unquote) [https://github.com/Particular/NServiceBus.SqlServer/blob/master/src/NServiceBus.SqlServer/Addressing/NameHelper.cs#L17] logic in SqlServer transport.

DavidBoike: :+1: I meant it as a curiosity more than anything, no need to change.

src/NServiceBus.NHibernate.Tests/TimeoutPersister/When_checking_db_indexes.cs
```diff
@@ -0,0 +1,123 @@
...
+    using global::NHibernate.Tool.hbm2ddl;
+    using NUnit.Framework;
+    using TimeoutPersisters.NHibernate.Config;
+    using TimeoutPersisters.NHibernate.Installer;
+
+    class When_checking_db_indexes
+    {
+        SchemaExport schemaExport;
+
+        string dbSchemaName;
```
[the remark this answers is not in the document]
tmasternak: In this case, the reason why those are global variables is that we need them in `TearDown` method to clean-up. Otherwise, the clean-up would have to be duplicated in each method.

PR #365, opened 2019-07-25
Fix case in the test

SzymonPobiega:


WilliamBZA:
I don't understand how this could have broken the linux tests

PR #404, opened 2020-02-14
Bump to SqlTranspot 4.3.1 which still is doing message driven pubsub

danielmarbach:


andreasohlund:
What's the drawback of going to v6? we would still test the saga and outbox right?

danielmarbach:
I reconsidered v6 under the light that the intention of these tests is to use a transport that exercises all the available storages. That is only fulfilled with v4.x range. From that perspective there is no value on upgrading to v6 until we drop timeout and subscription storage from the persister

andreasohlund:
Not 100% sure I agree, we have other tests that verify those storages and one of the key aspects of the tests is to see that the transport transaction is floated to the saga and outbox via the transport transaction and synchronized storage session. If I had to choose I would prioritize making sure that still works on v6 compared to testing a combo that is now more and more unlikely to be used in production?

danielmarbach:
Happy to discuss it with you in a short call because I disagree with your not 100% agreeing with me ;)

PR #429, opened 2020-05-12
Update to NServiceBus Core 7.3

andreasohlund:


src/NServiceBus.NHibernate/NServiceBus.NHibernate.csproj
```diff
@@ -14,7 +14,7 @@
     <PackageReference Include="Fody" Version="6.1.1" PrivateAssets="All" />
     <PackageReference Include="Janitor.Fody" Version="1.8.0" PrivateAssets="All" />
     <PackageReference Include="Obsolete.Fody" Version="5.2.1" PrivateAssets="All" />
-    <PackageReference Include="NServiceBus" Version="[7.0.1, 8.0.0)" />
+    <PackageReference Include="NServiceBus" Version="7.3.0" />
```
danielmarbach, line 17: I think this should be

`[7.3.0, 8.0.0)`

adamralph: I'm not sure it should change at all. IIRC we only bump the minimum version if something from that new version is required for a change being made. The tests are being bumped to 7.3.0 to ensure that the package works with 7.3.0.

timbussmann: what @adamralph said.

danielmarbach: Again this really depends on what we are trying to achieve. If we depend on functionality of 7.3 in the persister then we would raise the lower bound. On this PR it seems we are not depending on anything in 7.3 and therefore the lower bound doesn't necessarily have to change I agree with @adamralph and @timbussmann.

The scenario might be different though if we want to release a minor of the persistence that bounds to 7.3 of the core to guide users with additional warnings and obsoletes that the gateway persistence type is phasing out

andreasohlund: Yea, missed that, will update

andreasohlund: We don't depend on anything new so I will make it `[7.0.0, 8.0.0)`

adamralph: @andreasohlund should it not be `[7.0.1, 8.0.0)`? I.e. unchanged?

> The scenario might be different though if we want to release a minor of the persistence that bounds to 7.3 of the core to guide users with additional warnings and obsoletes that the gateway persistence type is phasing out

@danielmarbach I'm not sure I understand this. If there are new obsolete in core 7.3.0, they will see those when upgrading to 7.3.0. If we add new obsolete in this downstream, they will see those when upgrading to the new downstream.

danielmarbach: @adamralph my point was a general point not specific to this PR. For example in 7.2 we introduced a special mode in the core that handles timeout migration. RabbitMQ needs that and therefore we bumped the lower bound.

andreasohlund: @adamralph correct, its been reverted

adamralph: @danielmarbach in that case the downstream needs a new feature in core, so that makes sense. The point you made about obsoletes seems different.

danielmarbach:
It seems that the gateway acceptance test sources fail now with 

```
                          C:\Users\buildagent\.nuget\packages\nservicebus.gateway.acceptancetests.sources\3.1.0\contentFiles\cs\any\NSB.Gateway.AcceptanceTests\GatewayTestSuiteConstraints.cs(3, 56): error CS0535: 'GatewayTestSuiteConstraints' does not implement interface member 'IGatewayTestSuiteConstraints.ConfigureDeduplicationStorage(string, EndpointConfiguration, RunSettings)'
07:41:04
                          C:\Users\buildagent\.nuget\packages\nservicebus.gateway.acceptancetests.sources\3.1.0\contentFiles\cs\any\NSB.Gateway.AcceptanceTests\GatewayTestSuiteConstraints.cs(3, 56): error CS0535: 'GatewayTestSuiteConstraints' does not implement interface member 'IGatewayTestSuiteConstraints.Cleanup()'
07:41:04
                          C:\Users\buildagent\.nuget\packages\nservicebus.gateway.acceptancetests.sources\3.1.0\contentFiles\cs\any\NSB.Gateway.AcceptanceTests\When_using_legacy_persistence.cs(90, 42): error CS0618: 'IDeduplicateMessages' is obsolete: 'Gateway persistence has been moved to the NServiceBus.Gateway dedicated package. Will be treated as an error from version 8.0.0. Will be removed in version 9.0.0.'
```

see for example https://github.com/Particular/NServiceBus.Gateway/blob/master/src/NServiceBus.Gateway.AcceptanceTests/When_using_legacy_persistence.cs#L90

so this means the gateway ATT sources should be updated first as a pre-release and then be pulled in probably

andreasohlund:
@timbussmann I've managed to clean up the settings backdoor usage so this one should be good for a final review

PR #438, opened 2020-06-15
Pessimistic mode for the outbox

SzymonPobiega:
- Introduces an abstraction for outbox transaction with `Begin` and `Complete` operations
- Existing outbox transaction implementation renamed to _optimistic_
- New _pessimistic_ implementation
  - Insert empty `OutboxRecord` in `Begin`
  - Update `OutboxRecord` with outgoing messages in `Complete`
- `Begin` and `Complete` execute always in the same DB transaction to outside observers still see it as simple `insert`
- Limits the likelihood of applying non-transactional side effects multiple times in case of duplicate messages


SzymonPobiega:
Replaced by #439

PR #439, opened 2020-06-16
Outbox improvements

SzymonPobiega:
Replaces #438 (includes that one and also transaction scope support)

- Introduce pessimistic mode
- Introduce transaction scope support


src/NServiceBus.NHibernate/Outbox/PessimisticOutboxBehavior.cs
```diff
@@ -0,0 +1,29 @@
...
+                MessageId = endpointQualifiedMessageId,
+                Dispatched = false,
+                TransportOperations = null
+            };
+            return session.SaveAsync(outboxRecord);
+        }
+
+        public override Task Complete(string endpointQualifiedMessageId, ISession session, OutboxMessage outboxMessage, ContextBag context)
+        {
+            var queryString = $"update {typeof(TEntity).Name} set TransportOperations = :ops where MessageId = :messageid";
```
ramonsmits, line 22: @SzymonPobiega The entity was already saved and from that moment on I think it is tracked so why not do `ISession.Flush` here? Probably Flush is already called somewhere else so this seems unneeded.

SzymonPobiega: Nope, it does not work. This is the only way I was able to force NH to actually hit the database here. Neither Flush nor explicitly calling `Lock` helped.

SzymonPobiega:
> Pessimistic outbox would not work with snapshot isolation level. Maybe add a check to ensure the customer isn't doing this?

That would be a good idea but I think this is too complex to get right. I think it will be enough to mention in the doco. Users need to enable pessimistic mode anyway...

src/NServiceBus.NHibernate/Outbox/PessimisticOutboxBehavior.cs
ramonsmits: @SzymonPobiega That is strange, it is the same nhibernate session right? I just checked, the entity is tracked once Save has been called and any changes on it will result in an UPDATE after any Flush or Commit. Don't get fooled by `show_sql`, as the UPDATE does happen, it just isn't logged.

SzymonPobiega: Hmmm. Ok. I'll try to go through it with a profiler tomorrow. Why would `show_sql` not show the updates?

ramonsmits: @SzymonPobiega `show_sql` only works reliably if you disable batching `adonet.batch_size=0`.

SzymonPobiega: @ramonsmits I replaced this code with

```
record.TransportOperations = ConvertOperations(outboxMessage);
return session.FlushAsync();
```

and checked in the profiler and it does not issue an `update` so leaving the HQL. One we release this, feel free to created a PR to fix it if you find a way to force NH to flush the changes.

SzymonPobiega:
@tmasternak @dvdstelt can you review this one?

PR #449, opened 2020-06-25
Fix session lifecycle master

SzymonPobiega:


SzymonPobiega:
@tmasternak @dvdstelt can you approve? That's the backporting of our changes to master

PR #572, opened 2021-03-15
NHibernate: CancellationToken support + Core Alpha 910

DavidBoike:


andreasohlund (approved):
Should https://github.com/Particular/NServiceBus.NHibernate/blob/master/src/NServiceBus.NHibernate/SynchronizedStorage/INHibernateStorageSession.cs#L21 be made to accept a token as well?

src/NServiceBus.NHibernate.Tests/ApprovalFiles/APIApprovals.Approve.approved.txt
```diff
@@ -32,7 +32,9 @@ namespace NServiceBus
     public interface INHibernateStorageSession
     {
         NHibernate.ISession Session { get; }
+        [System.Obsolete("Use the overload that supports cancellation. Will be removed in version 10.0.0.", true)]
```
adamralph, line 35: What does it mean to obsolete an interface method? You still have to implement it.

DavidBoike: Yeah, but that's on us in the internal class implementations, right? It's a user uses the API of this interface, the user doesn't implement their own.

See https://github.com/Particular/NServiceBus.NHibernate/pull/572/files#diff-3e2f6e51b8b2efa9db6d34ab7098dde40deb15d2cdad551b17af2cb74c44826fR38-R46 for example.

adamralph: I see, so it's one of those public interfaces which someone could implement if they wanted to, but they'd have no use for that implementation, since they can't pass it into the API anywhere?

In that case, this obsoletion makes sense.

BTW - are the obsoletions on the internal types (e.g. the one you linked to) redundant? Since no-one will ever see them?

DavidBoike: They should help us to find and destroy them later on, since otherwise an internal class would be totally fine having an extraneous method that's not mentioned by the interface.

adamralph (approved):

PR #692, opened 2021-12-01
Migrate to GHA

helenktsai:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,102 @@
...
+        with:
+          dotnet-version: 6.0.x
+      - name: Setup .NET 5 
+        uses: actions/setup-dotnet@v1.9.0
+        with:
+          dotnet-version: 5.0.x          
+      - name: Setup .NET Core 3.1 
+        uses: actions/setup-dotnet@v1.9.0
+        with:
+          dotnet-version: 3.1.x
```
DavidBoike, line 63: Since you don't need net6.0 (yet) I wonder if switching to this would have any effect on the build problem you're getting on the self-hosted agent: (This combined versions stuff is possible since setup-dotnet 1.9.0)
```suggestion
      - name: Setup .NET SDK
        uses: actions/setup-dotnet@v1.9.0
        with:
          dotnet-version: |
            5.0.x
            3.1.x
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,95 @@
...
+        include:
+          - os: windows-2019
+            name: Windows-Sql
+          - os: ubuntu-20.04
+            name: Linux
+          - os: [self-hosted, windows, oracle]
+            name: Windows-Oracle
+      fail-fast: false
+    steps:
+      - name: Check for secrets
```
kbaley, line 36: Do we need this step if nothing in the build requires secrets?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,95 @@
...
+  pull_request_target:
+      branches-ignore:
+        - release-[0-7].*
+  workflow_dispatch:
+env:
+  DOTNET_NOLOGO: true
+jobs:
+  build:
+    if:
+      (github.event_name == 'pull_request_target' && github.event.pull_request.user.login == 'dependabot[bot]') ||
```
kbaley, line 20: Are these checks still required if this build doesn't rely on secrets?

.github/workflows/release.yml
```diff
@@ -0,0 +1,97 @@
+name: Release
+on:
+  push:
+    tags:
+      - '([8-9]|[1-9][0-9]).[0-9]+.[0-9]+'
```
kbaley, line 5: Not sure if this fancy syntax is better than an explicit extra line for builds higher than 9.x.x

kbaley (approved):

PR #744, opened 2022-03-04
update to core beta 1

HEskandari:


src/NServiceBus.NHibernate.Tests/SagaPersister/MySaga.cs
```diff
@@ -10,10 +10,8 @@ protected override void ConfigureHowToFindSaga(SagaPropertyMapper<MySagaData> ma
         }
     }
 
-    public class MySagaData : IContainSagaData
+    public class MySagaData : ContainSagaData
     {
-        public virtual Guid Id { get; set; }
-        public virtual string OriginalMessageId { get; set; }
-        public virtual string Originator { get; set; }
+        public Guid SagaId { get; set; }
```
timbussmann, line 15: should this be some `CorrelationId` instead of `SagaId` ? `SagaId` seems a bit confusing given that `ContainsSagaData` provides an `Id` property that is representing the saga id.

HEskandari: The analyzer guard against correlating with ID property derived from `IContainSagaData` . The ID property is for internal purposes only and shouldn't be used for correlation (at least that was the assumption under which we have created those analyzer rules). Even when inheriting from `ContainSagaData` (class instead of interface), we have hidden these 3 properties from the user (you do saga dot and you don't see these options in the intellisense).

HEskandari: Updated message/saga Ids as we discussed to `CorrelationId` to be consistent with the core.

timbussmann (approved):

mikeminutillo (approved):

PR #748, opened 2022-03-10
Set variable using powershell

timbussmann:
It looks like setting the oracle connection string for the linux builds wasn't set using the correct powershell syntax, making the connection string not set and therefore all oracle tests being ignored on the linux build.


DavidBoike (approved):
Yeah why can't pwsh just act like bash there?

DavidBoike:
done

PR #962, opened 2023-03-16
[Spike] Additional outbox+saga persistence tests

timbussmann:
Testing whether NH would be able to run these tests before considering adding them to the core persistence tests


danielmarbach:
Seems to fail with a key not found. Probably just some missing additional context data that needs to be present now that we might be able to add to the setup

timbussmann:
closing this as the test will be added via https://github.com/Particular/NServiceBus/pull/6706

PR #1005, opened 2023-06-16
Bind the storage session so that it doesn't throw when not available

danielmarbach:


src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateStorageSession.cs
```diff
@@ -56,7 +56,7 @@ protected override void Setup(FeatureConfigurationContext context)
             });
 
             context.Services.AddScoped<ICompletableSynchronizedStorageSession, NHibernateSynchronizedStorageSession>();
-            context.Services.AddScoped(sb => sb.GetRequiredService<ICompletableSynchronizedStorageSession>().StorageSession());
+            context.Services.AddScoped(sp => (sp.GetService<ISynchronizedStorageSession>() as INHibernateStorageSession)!);
```
timbussmann, line 59: If I understand https://github.com/Particular/NServiceBus.NHibernate/blob/master/src/NServiceBus.NHibernate/SynchronizedStorageSessionExtensions.cs#LL27C25-L27C51 correctly, the `ICompletableSynchronizedStorageSession` implementation doesn't directly implement `INHibernateStorageSession` but `INHibernateStorageSession` instead which is accessed to resolve the instance. The casting here would try to cast the session provider into a session instead.

danielmarbach: Ah yeah I remember now. Will fix

src/NServiceBus.NHibernate/SynchronizedStorage/NHibernateStorageSession.cs
```diff
@@ -56,7 +56,7 @@ protected override void Setup(FeatureConfigurationContext context)
             });
 
             context.Services.AddScoped<ICompletableSynchronizedStorageSession, NHibernateSynchronizedStorageSession>();
-            context.Services.AddScoped(sb => sb.GetRequiredService<ICompletableSynchronizedStorageSession>().StorageSession());
+            context.Services.AddScoped(sp => (sp.GetService<ISynchronizedStorageSession>() as INHibernateStorageSessionProvider)?.InternalSession);
```
timbussmann, line 59: since `ISynchronizedStorageSession` is only forwarding to `ICompletableSynchronizedStorageSession` anyway, is there a particular reason you went with `ISynchronizedStorageSession`?

danielmarbach: I did it that way because I was thinking from the binding perspective the completable version has methods that we are never interested in. That's how I did it for all others too. I'd be good either way. Would just have to send follow up PR for the already merged PRs

timbussmann: since the persisters are registering `ICompletableSynchronizedStorageSession` themselves, I think it might make a tiny bit more sense to use that interface so it's easier to understand what's being resolved without requiring to understand that core provides the extra registration for `ISynchronizedStorageSession` (which just forwards again). Feels more cohesive overall, but it's a nitpick tbh.

danielmarbach: How about we merge this one first ;)? Since we haven't caught this in the other already merged PRs if you help me I'm more than happy to revert using the other interface since my argument is also of the same fluffiness as yours. I can totally see both angles :D

timbussmann (approved):

PR #1043, opened 2023-09-12
Prepare repo for .NET 8

andreasohlund:
Downstream follow-up to [preparing for .NET 8](https://github.com/Particular/NServiceBus/pull/6810).


kentdr (approved):

src/NServiceBus.NHibernate/Outbox/NSBDateTimeType.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace NServiceBus.Deduplication.NHibernate.Config
+﻿namespace NServiceBus.Outbox.NHibernate
```
bording, line 1: Is there any migration concern around this namespace changing since this type is referenced in `OutboxRecordMapping`? It's not clear to be what `OutboxRecordMapping` is doing, so I wanted to make sure the type's namespace wouldn't be baked into some database table or something like that.

src/NServiceBus.NHibernate.Tests/Persistence/NHibernateConfigurationBuilderTests.cs
```diff
@@ -130,120 +125,5 @@ public void Should_skip_settings_that_are_not_for_persistence()
 
             CollectionAssert.IsSubsetOf(expected, builder.Build().Configuration.Properties);
         }
-
-#if NETFRAMEWORK
```
bording, line 134: Since these tests are doing things with custom AppDomains, I understand why you removed them, but I'm wondering these are valid scenarios we still need to be testing for in some other way instead.

I don't know enough of the specific details around NHibernate, but I assume it still uses these config files when running on .NET.

If that's true, then maybe we need to do something more here?

bording: Also, the Tests csproj is currently copying `hibernate.cfg.xml`and `Testing.config` to the output directory. I see `Testing.config` referenced in these tests. Are both of these files just for these removed tests? 

If so, we should clean those up also.

src/NServiceBus.NHibernate/Outbox/NSBDateTimeType.cs
andreasohlund: Good catch, I think we are fine but reverted the namespace just to be on the safe side

src/NServiceBus.NHibernate.Tests/Persistence/NHibernateConfigurationBuilderTests.cs
andreasohlund: I have removed the unused config files since there are tests already that checks that we can override

https://github.com/Particular/NServiceBus.NHibernate/blob/master/src/NServiceBus.NHibernate.Tests/Persistence/NHibernateConfigurationBuilderTests.cs#L65

and this just tests that we can override what NH read from its config file => I don't see enough value to justify us fiddling with customer exes that checks this

andreasohlund:
@bording ready for final review

bording (approved):

PR #1152, opened 2024-07-25
Update CI to add the Docker credentials

tamararivera:
Adds the credentials to the setup Oracle action

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,5 +56,8 @@ jobs:
         with:
           connection-string-name: OracleConnectionString
           tag: NHibernatePersistence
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -56,5 +56,8 @@ jobs:
         with:
           connection-string-name: OracleConnectionString
           tag: NHibernatePersistence
+          registry-login-server: index.docker.io
```
bording, line 59: Looking at other CI workflows, they haven't added this.

I think we can drop this line since we are using the default.

mikeminutillo (approved):

PR #1277, opened 2025-04-15
Add test for concurrent storage sessions

awright18:
Testing to see if storage session has the problem identified in https://github.com/Particular/NServiceBus.Persistence.Sql/issues/1235


awright18:
The test was completed

PR #1287, opened 2025-06-12
Fix race condition in RepeatedFailuresOverTimeCircuitBreaker

hsvotwa:
A port of https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1066

> Serializing the arm/disarm actions is not enough. The state transition preceding the arm/disarm action must be captured in the lock as well, otherwise the state transitions could happen properly but then the arm/disarm actions could be executed by the processor in the wrong order, leaving the logic of the circuit breaker itself intact but the consequences of the arm/disarm actions in an incorrect state.

> I also added xdoc as a way to document that the arm/disarm actions are called within a lock, and to maybe pave the way for this to be used as a shared component.


WilliamBZA (approved):

afprtclr:
`license/cla` check got stuck and we should get it working. I close and reopen the PR in that case to retrigger it. Please check if that works for you.

afprtclr (approved):

WilliamBZA (approved):

PR #1289, opened 2025-06-12
Send failures could lock transport into single concurrency mode

hsvotwa:
Backport of:
-  #1287

Which fixes for the `release-10.0` branch:
- #1291


afprtclr (approved):

afprtclr (changes requested):
I think we lack the Dispose method in this PR. Can you double check please?

PR #1426, opened 2026-05-19
Make `ScriptGenerator<T>.GenerateScript` public

abparticular:
The [NHibernate multitenant sample](https://docs.particular.net/samples/multi-tenant/nhibernate/) uses a public, but as yet undocumented API.

It is also misusing the API since it calls the `GenerateOutboxScript(Type)` method to generate the SQL script for the `Order` mapping, but this mapping is unrelated to the outbox.

There is a `GenerateScript(Type)` method that would be more suitable as a general script generator, but up until now it has not been public.

Refer to https://docs.particular.net/persistence/nhibernate/#generating-scripts-for-deployment-use-scriptgeneratort-api

### DIFF
--- a/src/NServiceBus.NHibernate/ScriptGenerator.cs
+++ b/src/NServiceBus.NHibernate/ScriptGenerator.cs
@@ -67,6 +67,8 @@
         /// <returns></returns>
         public static string GenerateScript(Type mappingType)
         {
+            ArgumentNullException.ThrowIfNull(mappingType);
+
             var config = new Configuration();
             config.DataBaseIntegration(db => { db.Dialect<T>(); });
 
--- a/src/NServiceBus.NHibernate/ScriptGenerator.cs
+++ b/src/NServiceBus.NHibernate/ScriptGenerator.cs
@@ -63,7 +63,7 @@
         /// <summary>
         /// Generates the table creation script for a mapping class.
         /// </summary>
-        /// <param name="mappingType"></param>
+        /// <param name="mappingType">The NHibernate mapping class.</param>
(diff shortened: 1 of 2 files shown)


src/NServiceBus.NHibernate/ScriptGenerator.cs
```diff
@@ -66,7 +60,12 @@ public static string GenerateSagaScript<TSaga>(Func<Type, string> tableNamingCon
             return GenerateScript(config);
         }
 
-        static string GenerateScript(Type mappingType)
+        /// <summary>
+        /// Generates the table creation script for a mapping class.
+        /// </summary>
+        /// <param name="mappingType"></param>
```
jasontaylordev:
```suggestion
        /// <param name="mappingType">The NHibernate mapping class.</param>
```

src/NServiceBus.NHibernate/ScriptGenerator.cs
```diff
@@ -66,7 +60,12 @@ public static string GenerateSagaScript<TSaga>(Func<Type, string> tableNamingCon
...
+        /// <returns></returns>
+        public static string GenerateScript(Type mappingType)
         {
             var config = new Configuration();
```
jasontaylordev, line 72: Consider adding a null guard here so callers get a clear `ArgumentNullException` with the parameter name rather than an opaque NHibernate internal exception.

```suggestion
            ArgumentNullException.ThrowIfNull(mappingType);

            var config = new Configuration();
```

jasontaylordev (approved):

